PR #29, opened 2017-05-20
Add blacklist role management

### LINKED ISSUE #28: Suggestion: User Blacklist
* Maintain a local list of users that are not allowed to use the bot. Alternatively check for a configurable role that disallows use of  the bot.
 * DJ and Admin status should override this blacklist.

james7132:
This is to resolve #28.
* Add a new role to Settings for blacklist
* Users on with this role cannot use any of the bot's commands
* DJs and Admins with the role can still use the bot's commands
* A `setblacklist` command was added to configure this at runtime.

I have not 100% tested this functionality yet due to lack of a fully set up development environment. Please notify me of any breakages that are caused by this change.

TODO: Instead of creating multiple commands for each role used, utilize a enum instead and pass it as a parameter.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/Bot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/Bot.java
@@ -52,7 +52,7 @@ import org.json.JSONObject;
  * @author John Grosh <john.a.grosh@gmail.com>
  */
 public class Bot extends ListenerAdapter {
-    
+
     private final HashMap<String,Settings> settings;
     private final AudioPlayerManager manager;
     private final EventWaiter waiter;
@@ -60,8 +60,22 @@ public class Bot extends ListenerAdapter {
     private JDA jda;
     private GUI gui;
     //private GuildsPanel panel;
-    public final Category MUSIC = new Category("Music");
-    public final Category DJ = new Category("DJ", event -> 
+    public final Category MUSIC = new Category("Music", event ->
+    {
+        if(event.getAuthor().getId().equals(event.getClient().getOwnerId()))
+            return true;
+        if(event.getGuild() == null)
+            return true;
+        if(event.getMember().hasPermission(Permission.MANAGE_SERVER))
+            return true;
+        List<Role> roles = event.getMember().getRoles();
+        Role dj = event.getGuild().getRoleById(getSettings(event.getGuild()).getDjRoleId());
+        if (roles.contains(dj))
+          return true;
+        Role blacklist = event.getGuild().getRoleById(getSettings(event.getGuild()).getBlacklistRoleId());
+        return !roles.contains(roles);
+    });
+    public final Category DJ = new Category("DJ", event ->
     {
         if(event.getAuthor().getId().equals(event.getClient().getOwnerId()))
             return true;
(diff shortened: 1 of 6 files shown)


src/main/java/com/jagrosh/jmusicbot/Bot.java
```diff
@@ -52,38 +52,52 @@
...
     private final EventWaiter waiter;
     private final ScheduledExecutorService threadpool;
     private JDA jda;
     private GUI gui;
     //private GuildsPanel panel;
-    public final Category MUSIC = new Category("Music");
-    public final Category DJ = new Category("DJ", event -> 
+    public final Category MUSIC = new Category("Music", event ->
+    {
+        if(event.getAuthor().getId().equals(event.getClient().getOwnerId()))
```
james7132: These share a lot of common checks with the others. It may be good to factor this out as common functionality.

jagrosh:
I already responded in the [issue](https://github.com/jagrosh/MusicBot/issues/28), but I don't think this type of feature is right for the bot (besides the fact that most of this won't compile). These kinds of configurations are a "slippery slope." First you add the ability to add a single role to the blacklist, then people want to be able to add multiple roles. After that, people want to assign different roles to different commands, blacklisting and allowing, etc etc. You eventually end up with an extremely complicated permissions system that most users will never use. The purpose of this bot is to be easy to use "out of the box," and I think it would be best if native Discord permissions were used. 

Besides, you can already blacklist a role or individual users. Simply set a text channel for the bot (`settc` command), and prevent that role or users from sending messages in that text channel. Or, set a voice channel for the bot (`setvc` command), and prevent that role or users from joining the voice channel.

If you would like to add this specific feature to your own copy, feel free to do so, but at present, I don't intend to add it. At some point in the future, I _do_ plan to add a bot-owner-controlled user blacklist that contains a list of user IDs that cannot use the bot at all, anywhere (abuse prevention). This, and other planned features, are outlined [here](https://github.com/jagrosh/MusicBot/projects/1)

james7132:
If there are compilation/logic issues, I'm fully willing to help resolve them. However, I fail to see a simple addition of a blacklist role leads to such a undesirable situation. 

To address the mentioned slippery slope situations, the supposed issue of "people want to be able to add multiple roles" shouldn't even be a problem since users can have more than one role. This is a reasonable situation to say no to, since there is a reasonable native alternative without side effects. The others are likewise similarly unreasonable feature requests, as they don't really add that much value to the bot as a whole, nor really appeal to the general use case.. 

This pull request, however, does. Per-server blacklists of users is important for server moderation, particularly for large servers where trolls and abuse abound. In our use case, the user who owns the bot is also relatively non-responsive, so a bot-owner curated blacklist would also be insufficient for cases that require fast response times. which this PR resolves, as adding a role can be done by anyone with moderator privileges, not just the bot owner. 

Presenting a solution such as blocking a users from using a channel outright, while effective, can also have side effects, namely blocking said users from any actual conversation in those channels: tying to manage this via a native Discord permission is that there are unintentional side effects when those permissions are removed.  Not to question your decisions here, but the DJ role configuration already exemplifies this. The bot could simply just check for `Manage Server`, but it also has a configurable role override: removing Manage Server from a user has much stronger implications than simply disallowing higher level of control over a bot. The blacklist role is a reasonable compromise: no managed data on the bot's end beyond the role ID, it uses a native Discord structure, and requires a simple one-time configuration. It makes no significant impact if left unconfigured, sticking to the easy to use "out of the box" ideal. 

If you still do not wish to add this as a feature, I'll set up a hard fork of the bot.

PhazonicRidley:
James, i would be interested in seeing ur version of the bot, a blacklist command is a great idea!

jagrosh:
Closing as this has been untouched for months

pbanj:
ya a blacklist feature would be pretty handy

alsoGAMER:
bump

PR #101, opened 2017-11-19
Hope to increase the language

henry753951:
Hope to increase the language 
Or let the user translate


jagrosh:
I'm just going to assume that this PR was an accident

PR #227, opened 2019-05-12
Play Command & Alias Improvements

seyahdoo:
### This pull request...
  - [ ] Improves an existing features

### Description
Play command will pause the music when given no arguments, 
Added 1 character aliases for some commands

#### added aliases:
- play -> p
- queue -> q
- stop -> s, leave, l
- remove -> r, d

### Purpose
This is done to make using this bot more fast by requiring less typing


jagrosh:
Closing this because neither of these are desired changes.
- Adding short aliases should not be something bot users are forced to use, and instead (if custom/shorter aliases are ever added) they should be able to set custom aliases for each command via the config file
- The pause command already pauses; the play command should not pause

PR #293, opened 2019-10-15
Add configurable maximum volume

### LINKED ISSUE #265: [Feature Request] Increase max volume input
Me and my friends loved to ear rape each other by setting the volume on our old music bot to some absurd number, but that music bot is now offline forever and no other bots seem to have this feature, so I'd love for you to implement unlimited volume input, at least as an optional setting perhaps, rather than capping at 150.

I'm hoping this would be a relatively easy fix, as I assume it is just changing the boundaries of the acceptable input for volume commands.

Thanks bro, your bot is awesome.

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR adds the option to configure the maximum volume in the config.txt-file. 

### Purpose
In the case of the author of #265, they want to use it to ear rape their friends just for fun.

### Relevant Issue(s)
This PR closes #265.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
+++ b/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
@@ -44,6 +44,7 @@ public class BotConfig
             successEmoji, warningEmoji, errorEmoji, loadingEmoji, searchingEmoji;
     private boolean stayInChannel, songInGame, npImages, updatealerts, useEval, dbots;
     private long owner, maxSeconds;
+    private int maxVolume;
     private OnlineStatus status;
     private Game game;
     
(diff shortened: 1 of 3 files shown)


jagrosh:
Frankly, I don't think this is something that should be configurable. The original limit was due to a limit within lavaplayer (used to be 0 to 150, now is 0 to 1000), and since this bot is designed to play music (not to play "earrape"), the bounds should be reasonable for that purpose. This is not an all-purpose bot; it is intended for music playback, and users intending to use a bot for non-music purposes should seek out a tool that is better aligned for their purpose.

More specific to this PR, the maximum should not be allowed to be set lower than 100, as playing at 100% volume means that some packets (almost all packets from YouTube) can be passed directly to Discord without decoding, which results in massively-reduced CPU usage.

Overall, I think the best solution here would be to just change the maximum to 200 (or maybe 300) instead of making the bot more complex by making it configurable.

PR #307, opened 2019-10-25
Add Aliases

MrSiliconGuy:
### This pull request...
  - [ ] Fixes a bug
  - [X] Improves an existing feature

### Description
Adds shortcut aliases to commonly used commands

### Purpose
Sometimes it is tiresome to always have to type !play for a command to work. Just like Rythm Bot, it might be nice to add a few shortcut aliases like !p for play, !s for skip, etc.

It's relatively easy to implement. You could even do it yourself instead of using this PR.

### Relevant Issue(s)
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,31 @@
 *.json
 *.txt
 nb*.xml
+/bin/
+*target*
+*.jar
+*.war
+*.ear
+*.class
+
+# eclipse specific git ignore
+*.pydevproject
+.project
+.metadata
+bin/**
+tmp/**
+tmp/**/*
+*.tmp
+*.bak
+*.swp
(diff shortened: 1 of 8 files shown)


jagrosh:
These should not be set by default, as they may conflict with other bots or are unwanted by some users. Additionally, these particular shortcuts used may not be the shortcuts that everyone wants. If this feature is added, it needs to be fully configurable (custom aliases for any command) from the config file, with the default setting being the current aliases.

jagrosh:
I'm going to close this for now. If you'd like to work on a config-file version of this, feel free to reopen this PR or open a new PR. Otherwise, I'll probably look into adding it myself fairly soon.

PR #308, opened 2019-10-25
Add configurable aliases for all commands

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR allows owners to define aliases for all commands in the config.txt file by giving `{command.name}` in the `aliases`-block the value of all aliases as a list of strings.
For example, if one wants to add the alias `s` to the `skip` command, then they should add the following inside the `aliases`-block in the config.txt:
`skip = [ s ]`

If one wants to add the aliases `yts` and `find` to the `search` command, then they should add the following inside the `aliases`-block in the config.txt:
`search = [ yts, find ]`

All currently existing aliases have been moved to `reference.conf`.

#### Current issues:
A vague error is being written to the console if one uses an alias that is already being used by another alias or command:
```
$ java -Dnogui=true -jar JMusicBot.jar
[INFO] [Startup]: Loaded config from MusicBot\config.txt
[ERROR] [JMusicBot]: Some aspect of the configuration is invalid: java.lang.IllegalArgumentException: Command added has a name or alias that has already been indexed: "v"!
Config Location: MusicBot\config.txt
$
```
In this case, the following conflict has happened (config.txt):
```
aliases {
  volume = [ v ]
  skip = [ v ]
}
```
`volume` and `skip` both have the aliases `v` defined and causes the MusicBot to crash when trying to start it.

### Purpose
Some owners may want to add one or two-character prefixes, such as `p` for `play`, `s` for `skip`, `fs` for `forceskip` and so on.

### Relevant Issue(s)
This PR is inspired by #307.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -152,6 +152,7 @@ aliases {
   shuffle = []
   skip = [ voteskip ]
 
+
   // Admin commands
 
   setdj = []
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -124,64 +124,66 @@ lyrics.default = "A-Z Lyrics"
 
 
 // These settings allow you to configure custom aliases for all commands.
-// Multiple aliases may be given, separated by spaces.
+// Multiple aliases may be given, separated by commas.
 //
 // Example 1: Giving command "play" the alias "p":
-// alias.play = "p"
+// play = [ p ]
 //
 // Example 2: Giving command "search" the aliases "yts" and "find":
-// alias.search = "yts find"
+// search = [ yts, find ]
 
-// General commands
+aliases {
 
-alias.settings = "status"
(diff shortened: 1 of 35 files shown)


MichailiK:
I just realized my last commit's "... & Add missing newline" is undescriptive: I added a missing newline in the `reference.conf`.

src/main/resources/reference.conf
```diff
@@ -123,6 +123,69 @@ updatealerts=true
...
+  debug = []
+  eval = []
+  playlist = []
+  setavatar = []
+  setgame = []
+  setname = []
+  setstatus = []
+  shutdown = []
+}
+
```
jagrosh: I'm still not sure I want to list every command here; something being not included should just default to an empty alias list. That being said, this is fine for now; I might just edit the example config or adjust this later.

PR #344, opened 2020-01-22
Add seek command

whew-inc:
### This pull request...
  - [ ] Fixes a bug

### Description
Allows DJs and user that added the track to perform a seek on a playing track.

### Purpose
Example: Going to a specific timestamp in a song compilation.

### Relevant Issue(s)
No relevant issue AFAIK, only a planned feature:
https://github.com/jagrosh/MusicBot/projects/1#card-1542746

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -109,9 +109,9 @@ public class JMusicBot
                         new RemoveCmd(bot),
                         new SearchCmd(bot),
                         new SCSearchCmd(bot),
+                        new SeekCmd(bot),
                         new ShuffleCmd(bot),
                         new SkipCmd(bot),
-                        new SeekCmd(bot),
 
                         new ForceRemoveCmd(bot),
                         new ForceskipCmd(bot),
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -111,6 +111,7 @@ public class JMusicBot
                         new SCSearchCmd(bot),
                         new ShuffleCmd(bot),
                         new SkipCmd(bot),
+                        new SeekCmd(bot),
 
                         new ForceRemoveCmd(bot),
                         new ForceskipCmd(bot),
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -67,7 +67,7 @@ public class SeekCmd extends MusicCommand
         int minutes = 0;
         int hours = 0;
 
-        if (Pattern.matches("^([0-9]\\d):([0-5]\\d):([0-5]\\d)$", args))
+        if (Pattern.matches("^(\\d\\d):([0-5]\\d):([0-5]\\d)$", args))
         {
             hours = Integer.parseInt(args.substring(0, 2));
             minutes = Integer.parseInt(args.substring(3, 5));
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -32,13 +32,11 @@ public class SeekCmd extends MusicCommand {
             AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
             Settings settings = event.getClient().getSettingsFor(event.getGuild());
 
-            if (!event.getMember().hasPermission(Permission.MANAGE_SERVER)) {
-                if (!event.getMember().getRoles().contains(settings.getRole(event.getGuild()))) {
-                    if (currentTrack.getUserData(Long.class) != event.getAuthor().getIdLong()) {
-                        event.replyError("You cannot seek **" + currentTrack.getInfo().title + "** because you didn't add it!");
-                        return;
-                    }
-                }
+            if (!event.getMember().hasPermission(Permission.MANAGE_SERVER)
+                    && !event.getMember().getRoles().contains(settings.getRole(event.getGuild()))
+                    && currentTrack.getUserData(Long.class) != event.getAuthor().getIdLong()) {
+                event.replyError("You cannot seek **" + currentTrack.getInfo().title + "** because you didn't add it!");
+                return;
             }
 
             String args = event.getArgs();
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -23,8 +23,6 @@ import com.jagrosh.jmusicbot.commands.MusicCommand;
 import com.jagrosh.jmusicbot.utils.TimeUtil;
 import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
 
-import java.util.regex.Pattern;
-
 
 /**
  * @author Whew., Inc.
@@ -36,7 +34,7 @@ public class SeekCmd extends MusicCommand
         super(bot);
         this.name = "seek";
         this.help = "seeks the current song";
-        this.arguments = "<HH:MM:SS>|<MM:SS>|<SS>";
+        this.arguments = "[+ | -] <HH:MM:SS | MM:SS | SS>";
         this.aliases = bot.getConfig().getAliases(this.name);
         this.beListening = true;
         this.bePlaying = true;
@@ -46,55 +44,36 @@ public class SeekCmd extends MusicCommand
     public void doCommand(CommandEvent event)
     {
         AudioHandler handler = (AudioHandler) event.getGuild().getAudioManager().getSendingHandler();
-        if (!handler.getPlayer().getPlayingTrack().isSeekable())
+        AudioTrack playingTrack = handler.getPlayer().getPlayingTrack();
+        if (!playingTrack.isSeekable())
         {
             event.replyError("This track is not seekable.");
             return;
         }
 
-        AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
-        if (!DJCommand.checkDJPermission(event))
+
+        if (!DJCommand.checkDJPermission(event) && playingTrack.getUserData(Long.class) != event.getAuthor().getIdLong())
         {
-            event.replyError("You cannot seek **" + currentTrack.getInfo().title + "** because you didn't add it!");
+            event.replyError("You cannot seek **" + playingTrack.getInfo().title + "** because you didn't add it!");
             return;
         }
 
         String args = event.getArgs();
-        long track_duration = handler.getPlayer().getPlayingTrack().getDuration();
-        int seek_milliseconds = 0;
-        int seconds;
-        int minutes = 0;
-        int hours = 0;
-
-        if (Pattern.matches("^(\\d\\d):([0-5]\\d):([0-5]\\d)$", args))
+        Long seek_milliseconds = TimeUtil.parseTime(args);
+        if (seek_milliseconds == null)
         {
-            hours = Integer.parseInt(args.substring(0, 2));
-            minutes = Integer.parseInt(args.substring(3, 5));
-            seconds = Integer.parseInt(args.substring(6));
-        }
-        else if (Pattern.matches("^([0-5]\\d):([0-5]\\d)$", args))
-        {
-            minutes = Integer.parseInt(args.substring(0, 2));
-            seconds = Integer.parseInt(args.substring(3, 5));
-        }
-        else if (Pattern.matches("^([0-5]\\d)$", args))
-        {
-            seconds = Integer.parseInt(args.substring(0, 2));
-        }
-        else
-        {
-            event.replyError("Invalid seek!");
+            event.replyError("Invalid seek! Expected format:" + arguments);
             return;
         }
 
-        seek_milliseconds += hours * 3600000 + minutes * 60000 + seconds * 1000;
+        long track_duration = playingTrack.getDuration();
         if (seek_milliseconds > track_duration)
         {
             event.replyError("Current track (`" + TimeUtil.formatTime(track_duration) + "`) is not that long!");
             return;
         }
 
-        handler.getPlayer().getPlayingTrack().setPosition(seek_milliseconds);
-        event.replySuccess("Successfully seeked to `" + TimeUtil.formatTime(seek_milliseconds) + "`!");
+        playingTrack.setPosition(seek_milliseconds);
+        event.replySuccess("Successfully seeked to `" + TimeUtil.formatTime(seek_milliseconds) + "/" + TimeUtil.formatTime(playingTrack.getDuration()) + "`!");
     }
 }
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -66,21 +66,15 @@ public class SeekCmd extends MusicCommand
             return;
         }
 
-        long seek_milliseconds = seekTime.milliseconds;
         long currentPosition = playingTrack.getPosition();
-        long track_duration = playingTrack.getDuration();
+        long trackDuration = playingTrack.getDuration();
 
-        if (seekTime.relative != null)
+        long seekMilliseconds = seekTime.relative ? currentPosition + seekTime.milliseconds : seekTime.milliseconds;
+        if (seekMilliseconds > trackDuration)
         {
-            seek_milliseconds = seekTime.relative ? currentPosition + seekTime.milliseconds : currentPosition - seekTime.milliseconds;
-        }
-
-        if (seek_milliseconds > track_duration)
-        {
-            event.replyError("Cannot seek to `" + TimeUtil.formatTime(seek_milliseconds) + "` because the current track is `" + TimeUtil.formatTime(track_duration) + "` long!");
+            event.replyError("Cannot seek to `" + TimeUtil.formatTime(seekMilliseconds) + "` because the current track is `" + TimeUtil.formatTime(trackDuration) + "` long!");
             return;
-        }
-        else playingTrack.setPosition(seek_milliseconds);
+        } else playingTrack.setPosition(seekMilliseconds);
 
         event.replySuccess("Successfully seeked to `" + TimeUtil.formatTime(playingTrack.getPosition()) + "/" + TimeUtil.formatTime(playingTrack.getDuration()) + "`!");
     }
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -59,36 +59,25 @@ public class SeekCmd extends MusicCommand
         }
 
         String args = event.getArgs();
-        Boolean seek_relative = null; // seek forward or backward
-        char charRelative = args.charAt(0);
-        if (args.charAt(0) == '+' || args.charAt(0) == '-')
+        TimeUtil.SeekTime seekTime = TimeUtil.parseTime(args);
+        if (seekTime == null)
         {
-            args = args.substring(1);
-            seek_relative = charRelative == '+';
-
-        }
-        Long seek_milliseconds = TimeUtil.parseTime(args);
-        if (seek_milliseconds == null)
-        {
-            event.replyError("Invalid seek! Expected format:" + arguments);
+            event.replyError("Invalid seek! Expected format: " + arguments);
             return;
         }
 
-        long track_duration = playingTrack.getDuration();
+        long seek_milliseconds = seekTime.milliseconds;
         long currentPosition = playingTrack.getPosition();
-        if (seek_relative != null)
+        long track_duration = playingTrack.getDuration();
+
+        if (seekTime.relative != null)
         {
-            if ((seek_relative && (seek_milliseconds > (track_duration - currentPosition))) || (!seek_relative && (seek_milliseconds > currentPosition)))
-            {
-                event.replyError("You can't seek past the length of the track! (" + TimeUtil.formatTime(currentPosition) + "/" + TimeUtil.formatTime(track_duration) + ")");
-                return;
-            }
-            playingTrack.setPosition(seek_relative ? currentPosition + seek_milliseconds : currentPosition - seek_milliseconds);
+            seek_milliseconds = seekTime.relative ? currentPosition + seekTime.milliseconds : currentPosition - seekTime.milliseconds;
         }
 
-        else if (seek_milliseconds > track_duration)
+        if (seek_milliseconds > track_duration)
         {
-            event.replyError("Current track (`" + TimeUtil.formatTime(track_duration) + "`) is not that long!");
+            event.replyError("Cannot seek to `" + TimeUtil.formatTime(seek_milliseconds) + "` because the current track is `" + TimeUtil.formatTime(track_duration) + "` long!");
             return;
         }
         else playingTrack.setPosition(seek_milliseconds);
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -95,7 +95,7 @@ public class SeekCmd extends MusicCommand
         }
 
         handler.getPlayer().getPlayingTrack().setPosition(seek_milliseconds);
-        event.replySuccess("Successfully seeked!");
+        event.replySuccess("Successfully seeked to `" + FormatUtil.formatTime(seek_milliseconds) + "`!");
     }
 
 
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -20,6 +20,7 @@ import com.jagrosh.jmusicbot.Bot;
 import com.jagrosh.jmusicbot.audio.AudioHandler;
 import com.jagrosh.jmusicbot.commands.DJCommand;
 import com.jagrosh.jmusicbot.commands.MusicCommand;
+import com.jagrosh.jmusicbot.utils.FormatUtil;
 import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
 
 import java.util.regex.Pattern;
@@ -89,7 +90,7 @@ public class SeekCmd extends MusicCommand
         seek_milliseconds += hours * 3600000 + minutes * 60000 + seconds * 1000;
         if (seek_milliseconds > track_duration)
         {
-            event.replyError("Current track is not that long!");
+            event.replyError("Current track (`" + FormatUtil.formatTime(track_duration) + "`) is not that long!");
             return;
         }
 
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -33,8 +33,7 @@ public class SeekCmd extends MusicCommand
     {
         super(bot);
         this.name = "seek";
-        this.help = "seeks the current song" +
-                "\nExamples: `1:02:23` `+1:10` `-90`";
+        this.help = "seeks the current song";
         this.arguments = "[+ | -] <HH:MM:SS | MM:SS | SS>";
         this.aliases = bot.getConfig().getAliases(this.name);
         this.beListening = true;
@@ -63,7 +62,7 @@ public class SeekCmd extends MusicCommand
         TimeUtil.SeekTime seekTime = TimeUtil.parseTime(args);
         if (seekTime == null)
         {
-            event.replyError("Invalid seek! Expected format: " + arguments);
+            event.replyError("Invalid seek! Expected format: " + arguments + "\nExamples: `1:02:23` `+1:10` `-90`");
             return;
         }
 
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -18,6 +18,7 @@ package com.jagrosh.jmusicbot.commands.music;
 import com.jagrosh.jdautilities.command.CommandEvent;
 import com.jagrosh.jmusicbot.Bot;
 import com.jagrosh.jmusicbot.audio.AudioHandler;
+import com.jagrosh.jmusicbot.commands.DJCommand;
 import com.jagrosh.jmusicbot.commands.MusicCommand;
 import com.jagrosh.jmusicbot.settings.Settings;
 import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
@@ -53,11 +54,7 @@ public class SeekCmd extends MusicCommand
         }
 
         AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
-        Settings settings = event.getClient().getSettingsFor(event.getGuild());
-
-        if (!event.getMember().hasPermission(Permission.MANAGE_SERVER)
-                && !event.getMember().getRoles().contains(settings.getRole(event.getGuild()))
-                && currentTrack.getUserData(Long.class) != event.getAuthor().getIdLong())
+        if (!DJCommand.checkDJPermission(event))
         {
             event.replyError("You cannot seek **" + currentTrack.getInfo().title + "** because you didn't add it!");
             return;
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
@@ -46,61 +46,59 @@ public class SeekCmd extends MusicCommand
     public void doCommand(CommandEvent event)
     {
         AudioHandler handler = (AudioHandler) event.getGuild().getAudioManager().getSendingHandler();
-        if (handler.getPlayer().getPlayingTrack().isSeekable())
+        if (!handler.getPlayer().getPlayingTrack().isSeekable())
         {
-            AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
-            Settings settings = event.getClient().getSettingsFor(event.getGuild());
+            event.replyError("This track is not seekable.");
+            return;
+        }
 
-            if (!event.getMember().hasPermission(Permission.MANAGE_SERVER)
-                    && !event.getMember().getRoles().contains(settings.getRole(event.getGuild()))
-                    && currentTrack.getUserData(Long.class) != event.getAuthor().getIdLong())
-            {
-                event.replyError("You cannot seek **" + currentTrack.getInfo().title + "** because you didn't add it!");
-                return;
-            }
+        AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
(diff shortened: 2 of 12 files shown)


src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
```diff
@@ -111,6 +111,7 @@ public static void main(String[] args)
                         new SCSearchCmd(bot),
                         new ShuffleCmd(bot),
                         new SkipCmd(bot),
+                        new SeekCmd(bot),
```
jagrosh: Commands should be listed in alphabetical order, so this should be above the shuffle command

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
+package com.jagrosh.jmusicbot.commands.music;
```
jagrosh: This is missing the license header

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
+import net.dv8tion.jda.core.Permission;
+
+import java.util.regex.Pattern;
+
+
+/**
+ * @author Whew., Inc.
+ */
+public class SeekCmd extends MusicCommand {
```
jagrosh: Brackets should be on next line
```java
public class SeekCmd extends MusicCommand
{
...
}
```

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+        this.arguments = "<HH:MM:SS>|<MM:SS>|<SS>";
+        this.aliases = bot.getConfig().getAliases(this.name);
+        this.beListening = true;
+        this.bePlaying = true;
+    }
+
+    @Override
+    public void doCommand(CommandEvent event) {
+        AudioHandler handler = (AudioHandler) event.getGuild().getAudioManager().getSendingHandler();
+        if (handler.getPlayer().getPlayingTrack().isSeekable()) {
```
jagrosh: For easier readability, instead of structuring the if-statements like how you have it here, consider:
```java
if(!handler.getPlayer().getPlayingTrack().isSeekable())
{
    event.replyError("This track is not seekable.");
    return;
}
AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
...
```

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+            AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
+            Settings settings = event.getClient().getSettingsFor(event.getGuild());
+
+            if (!event.getMember().hasPermission(Permission.MANAGE_SERVER)
+                    && !event.getMember().getRoles().contains(settings.getRole(event.getGuild()))
```
jagrosh: This check should use `DJCommand#checkDJPermission` instead of these first two conditionals

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+                seconds = Integer.parseInt(args.substring(6));
+            } else if (Pattern.matches("^([0-5]\\d):([0-5]\\d)$", args)) {
+                minutes = Integer.parseInt(args.substring(0, 2));
+                seconds = Integer.parseInt(args.substring(3, 5));
+            } else if (Pattern.matches("^([0-5]\\d)$", args)) {
+                seconds = Integer.parseInt(args.substring(0, 2));
+            } else {
+                event.replyError("Invalid seek!");
+                return;
+            }
```
jagrosh: I'm not a huge fan of having this here; instead maybe it could go in a `private int parseTime(String)` method that returns a number of seconds.

Alternatively, move the logic here as well as the time-formatting methods in FormatUtil to a new TimeUtil to make it easier to see the consistency of input and output.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+
+            seek_milliseconds += hours * 3600000 + minutes * 60000 + seconds * 1000;
+            if (seek_milliseconds <= track_duration) {
+                handler.getPlayer().getPlayingTrack().setPosition(seek_milliseconds);
+                event.replySuccess("Successfully seeked!");
+            } else {
+                event.replyError("Current track is not that long!");
```
jagrosh: This should output the length of the track

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,74 @@
...
+                seconds = Integer.parseInt(args.substring(0, 2));
+            } else {
+                event.replyError("Invalid seek!");
+                return;
+            }
+
+            seek_milliseconds += hours * 3600000 + minutes * 60000 + seconds * 1000;
+            if (seek_milliseconds <= track_duration) {
+                handler.getPlayer().getPlayingTrack().setPosition(seek_milliseconds);
+                event.replySuccess("Successfully seeked!");
```
jagrosh: This should restate the point to which the track was seeked, reformatted in case the user had weird input (ie, and input of `01:95` should output that the track was seeked to `02:35`)

Whew-Inc: The current regex filter doesn't allow for weird input like that. Do you think we should allow such input, or keep it like it is now?

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
Whew-Inc: I went with the second option for now. It required some refactoring. Tell me what you think of it.

Whew-Inc:
How do you think we should deal with invalid input? Three options that came to mind are try/catch, optional and null.

Whew-Inc:
By the way sorry if this PR is messy, this is my first one. If you got any suggestions on how I can do it better, I'm open to them :)

jagrosh:
Since the time parsing will likely need to be able to handle both specific and relative times (`+`/`-`), a new class for the return type will likely be needed. Probably something like
```java
public static class SeekTime
{
    public final int seconds;
    public final boolean relative;
    /* constructor here */
}
```
> How do you think we should deal with invalid input in `parseTime`? Three options that came to mind are throw exception, optional and null.

Exception or null, I tend to use nulls but exceptions are useful if there are multiple fail-cases.

> By the way sorry if this PR is messy, this is my first one. If you got any suggestions on how I can do it better, I'm open to them :)

Generally it's better to keep similar changes in a single commit, but I can squash before merging so it's not a huge deal.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,100 @@
...
+    {
+        AudioHandler handler = (AudioHandler) event.getGuild().getAudioManager().getSendingHandler();
+        if (!handler.getPlayer().getPlayingTrack().isSeekable())
+        {
+            event.replyError("This track is not seekable.");
+            return;
+        }
+
+        AudioTrack currentTrack = handler.getPlayer().getPlayingTrack();
+        if (!DJCommand.checkDJPermission(event))
```
jagrosh: I think you accidentally removed the check that allowed people to seek their own tracks (this should check if the user added the track or they are a DJ)

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author Whew., Inc.
+ */
+public class SeekCmd extends MusicCommand
+{
+    public SeekCmd(Bot bot)
+    {
+        super(bot);
+        this.name = "seek";
+        this.help = "seeks the current song";
+        this.arguments = "<HH:MM:SS>|<MM:SS>|<SS>";
```
jagrosh: `|` between arguments implies a specific character, this should be `<[[HH:]MM:]SS>` or `<HH:MM:SS | MM:SS | SS>`. With the recommendation for how to parse relative times, I think this should probably be `[+ | -] <HH:MM:SS | MM:SS | SS>`

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,100 @@
...
+            minutes = Integer.parseInt(args.substring(0, 2));
+            seconds = Integer.parseInt(args.substring(3, 5));
+        }
+        else if (Pattern.matches("^([0-5]\\d)$", args))
+        {
+            seconds = Integer.parseInt(args.substring(0, 2));
+        }
+        else
+        {
+            event.replyError("Invalid seek!");
```
jagrosh: This might be a good place to reiterate the expected input format

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
Whew-Inc: You're right, whoops.

Whew-Inc:
I like the idea with the + and - before the timestamp. I'll do the forward/backward seeking in the coming days.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,98 @@
...
+
+        String args = event.getArgs();
+        Boolean seek_relative = null; // seek forward or backward
+        char charRelative = args.charAt(0);
+        if (args.charAt(0) == '+' || args.charAt(0) == '-')
+        {
+            args = args.substring(1);
+            seek_relative = charRelative == '+';
+
+        }
```
jagrosh: This should be part of the parseTime method, which should return a child class that contains a `boolean` and a `long`.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,98 @@
...
+                return;
+            }
+            playingTrack.setPosition(seek_relative ? currentPosition + seek_milliseconds : currentPosition - seek_milliseconds);
+        }
+
+        else if (seek_milliseconds > track_duration)
+        {
+            event.replyError("Current track (`" + TimeUtil.formatTime(track_duration) + "`) is not that long!");
+            return;
+        }
```
jagrosh: All of this can be simplified by calculating the position that the user is trying to seek to, and then determining if that position is within the bounds. An example error message might be **Cannot seek to `01:13` because the current track is `01:02` long!**. This avoids the need to have different messages for relative and direct seeks.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,98 @@
...
+        Boolean seek_relative = null; // seek forward or backward
+        char charRelative = args.charAt(0);
+        if (args.charAt(0) == '+' || args.charAt(0) == '-')
+        {
+            args = args.substring(1);
+            seek_relative = charRelative == '+';
+
+        }
+        Long seek_milliseconds = TimeUtil.parseTime(args);
+        if (seek_milliseconds == null)
```
jagrosh: This should be a static child class of TimeUtil

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,98 @@
...
+        {
+            event.replyError("Invalid seek! Expected format:" + arguments);
```
jagrosh: Needs a space and to be surrounded by code ticks

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,63 @@
...
+            return "LIVE";
+        long seconds = Math.round(duration/1000.0);
+        long hours = seconds/(60*60);
+        seconds %= 60*60;
+        long minutes = seconds/60;
+        seconds %= 60;
+        return (hours>0 ? hours+":" : "") + (minutes<10 ? "0"+minutes : minutes) + ":" + (seconds<10 ? "0"+seconds : seconds);
+    }
+
+    public static Long parseTime(String args)
```
jagrosh: This should return a ParsedTime (or similar name) that contains a `long` and a `boolean`; see one of my earlier comments for an example of this class.

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+package com.jagrosh.jmusicbot.utils;
+
+import java.util.regex.Pattern;
+
+public class TimeUtil
+{
+
+    public static String formatTime(long duration)
```
jagrosh: This and the other method here should probably have some documentation added to specify that they expect/return milliseconds

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
Whew-Inc: Yeah that might be better, less clutter too

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public static SeekTime parseTime(String args)
+    {
+        Boolean seek_relative = null; // seek forward or backward
+        char charRelative = args.charAt(0);
+        if (args.charAt(0) == '+' || args.charAt(0) == '-')
+        {
+            args = args.substring(1);
+            seek_relative = charRelative == '+';
+        }
```
jagrosh: I feel like this can be simplified to:
```java
if(args.length() == 0)
    return null;
boolean relative = false;
char first = args.charAt(0);
if(first == '+' || first == '-')
{
    args = args.substring(1);
    relative = true;
}
```
(also note that I changed `seek_relative` to `relative`; `seekRelative` would have been fine too, but don't use snake_case for variable names)

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        else if (Pattern.matches("^([0-5]\\d):([0-5]\\d)$", args))
+        {
+            minutes = Integer.parseInt(args.substring(0, 2));
+            seconds = Integer.parseInt(args.substring(3, 5));
+        }
+        else if (Pattern.matches("^([0-5]\\d)$", args))
+        {
+            seconds = Integer.parseInt(args.substring(0, 2));
+        }
+        else return null;
```
jagrosh: I think this regex should be relaxed just a bit; if someone wants to jump forward 90 seconds, they should be able to type `+90` and not be forced to type `+01:30`

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        return new SeekTime(seek_relative, hours * 3600000 + minutes * 60000 + seconds * 1000);
+    }
+
+
+    public static class SeekTime
+    {
+        public final long milliseconds;
+        public final Boolean relative;
+
+        public SeekTime(Boolean relative, long milliseconds)
```
jagrosh: This constructor should be private or protected (and use `boolean`, not `Boolean`)

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,87 @@
...
+
+        String args = event.getArgs();
+        TimeUtil.SeekTime seekTime = TimeUtil.parseTime(args);
+        if (seekTime == null)
+        {
+            event.replyError("Invalid seek! Expected format: " + arguments);
+            return;
+        }
+
+        long seek_milliseconds = seekTime.milliseconds;
```
jagrosh: This isn't really needed; seekTime already has the millisecond count as a public final variable

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,87 @@
...
+        long currentPosition = playingTrack.getPosition();
+        long track_duration = playingTrack.getDuration();
```
jagrosh: `trackDuration`

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
Whew-Inc: Whoops. Haven't worked with Java in a while, guess it's showing :P

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -53,11 +65,22 @@ else if (Pattern.matches("^([0-5]\\d)$", args))
...
+
+
+        return new SeekTime(seek_relative, hours * 3600000 + minutes * 60000 + seconds * 1000);
+    }
+
+
+    public static class SeekTime
+    {
+        public final long milliseconds;
+        public final Boolean relative;
```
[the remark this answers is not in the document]
Whew-Inc: I went with Boolean because the primitive type can't hold a null (for when we're not seeking relatively)

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
Whew-Inc: It's modified though, to calculate the final seek time.
https://github.com/Whew-Inc/MusicBot/blob/c6801ef3cf06375bc8378b533bf87caef6fd4984/src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java#L73-L76

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
Whew-Inc: This assumes we will always seek relatively though.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
jagrosh: Ah right, I missed that as it is further down. I'd move this down a few lines and change this to
```java
long seekMilliseconds = seekTime.relative ? currentPosition + seekTime.milliseconds : seekTime.milliseconds;
```

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
jagrosh: Also I missed this in my earlier review but don't forget to make the final millisecond value negative if `first == '-'`

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
Whew-Inc: Yeah, that would be easier.

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
jagrosh: > This assumes we will always seek relatively though.

No, we only set relative to true if it starts with `+` or `-`. By default, `relative` will be `false` and so we won't seek relative.

Whew-Inc:
I think we should allow timestamps without the preceding 0, like for example 1:00. Having to write 0s in a timestamp when it's not really needed is inefficient.

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,89 @@
...
+                timeUnitArray[timeUnitIndex] = Integer.parseUnsignedInt(timeUnit);
+            } catch (NumberFormatException e)
+            {
+                return null;
+            }
+            timeUnitIndex++;
+        }
+
+        long milliseconds = timeUnitArray[0] * 3600000 + timeUnitArray[1] * 60000 + timeUnitArray[2] * 1000;
+        if (relative && isSeekingBackwards) milliseconds = -milliseconds;
```
jagrosh: I think this can be simplified to:
```java
String[] timestampSplitArray = timestamp.split(":");
if(timestampSplitArray.length > 3 )
    return null;
int[] timeUnitArray = new int[3]; // hours, minutes, seconds
for(int index = 0; index < timestampSplitArray.length; index++)
{
    try
    {
        timeUnitArray[index + 3 - timestampSplitArray.length] = Integer.parseUnsignedInt(timestampSplitArray[index]);
    }
    catch (NumberFormatException e)
    {
        return null;
    }
}
long milliseconds = timeUnitArray[0] * 3600000 + timeUnitArray[1] * 60000 + timeUnitArray[2] * 1000;
milliseconds *= isSeekingBackwards ? -1 : 1;
```

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,81 @@
...
+        if (!playingTrack.isSeekable())
+        {
+            event.replyError("This track is not seekable.");
+            return;
+        }
+
+
+        if (!DJCommand.checkDJPermission(event) && playingTrack.getUserData(Long.class) != event.getAuthor().getIdLong())
+        {
+            event.replyError("You cannot seek **" + playingTrack.getInfo().title + "** because you didn't add it!");
```
Sanduhr32: I would change the detailed report of the error to something like "❌ You are missing permissions to seek tunes"

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        boolean isSeekingBackwards = false;
+        char first = timestamp.charAt(0);
+        if (first == '+' || first == '-')
+        {
+            relative = true;
+            isSeekingBackwards = first == '-';
+            timestamp = timestamp.substring(1);
+        }
+
+        String[] timestampSplitArray = timestamp.split(":");
```
Sanduhr32: Maybe change the regex to `:+` for also accepting double char typos?

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        return (hours>0 ? hours+":" : "") + (minutes<10 ? "0"+minutes : minutes) + ":" + (seconds<10 ? "0"+seconds : seconds);
+    }
+
+    /**
+     * @param args timestamp formatted as: [+ | -] &lt;HH:MM:SS | MM:SS | SS&gt;
+     * @return Time in milliseconds, negative if seeking backwards relatively
+     */
+    public static SeekTime parseTime(String args)
+    {
+        if (args.length() == 0 || args.length() > 8) return null;
```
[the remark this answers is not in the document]
Sanduhr32: I also believe that any amount of beginning zeros should be accepted. Beginning zeros should not interfere seeking. In fact, seeking half a second back would also be nice.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,81 @@
...
+
+        long currentPosition = playingTrack.getPosition();
+        long trackDuration = playingTrack.getDuration();
+
+        long seekMilliseconds = seekTime.relative ? currentPosition + seekTime.milliseconds : seekTime.milliseconds;
+        if (seekMilliseconds > trackDuration)
+        {
+            event.replyError("Cannot seek to `" + TimeUtil.formatTime(seekMilliseconds) + "` because the current track is `" + TimeUtil.formatTime(trackDuration) + "` long!");
+            return;
+        } else playingTrack.setPosition(seekMilliseconds);
```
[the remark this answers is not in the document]
Whew-Inc: Youtube, soundcloud, and mp3s work fine.
Twitch streams don't seem to be seekable, and clips fail to load.
Haven't tested other platforms/formats.

Anyway, I agree. It's probably a good idea to put it in a try catch block.
Perhaps reply with "`An error occurred when trying to seek!`", and print the error in the log?

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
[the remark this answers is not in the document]
Whew-Inc: > Although unlikely, a song or track could last hundreds of hours, in which case seeking to `105:12:15` would be impossible. I don't think the args length limit is needed here, as the integer parsing later will handle invalid inputs.

I agree.

> I also believe that any amount of beginning zeros should be accepted. Beginning zeros should not interfere seeking. In fact, seeking like 3 and half a seconds would also be nice as an example.

Seeking with decimal numbers would be nice, I agree. Coming to think of it, it might be even nicer if you could use any float number as argument. As long as it adheres to the x:x:x format, I think it would be fine.

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
Whew-Inc: Yes, this looks better. Though I think it might be best to check for a positive sign (+) at the start of the time unit. According to an excerpt from the `Integer#parseUnsignedInt` doc: 
_The characters in the string must all be decimal digits, except that the first character may be an an ASCII plus sign '+' ('\u002B')._

Although it probably wouldn't hurt, being able to seek with `++01:00` seems weird to me.
Aaand I just noticed the doc has a typo. `may be an an ASCII`. Whew.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
Whew-Inc: I think it's better to keep it like it is now. It's more consistent, since other commands use similar messages. (`RemoveCmd` has almost the exact same message)

src/main/java/com/jagrosh/jmusicbot/utils/TimeUtil.java
Whew-Inc: That would be handy, but I'm not sure if we should process input like `01::::00::01`.

Sanduhr32: Me neither but its handy and it would not cause errors on typos but it will error of you drop numerals when you want to simplify it. However `::40` might fail on it's own due to missing numerals.

Whew-Inc:
Seek with decimals, no timestamp separator `:` limit, lavaplayer seek error catching and some basic tests added.

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,88 @@
...
+ * @author Whew., Inc.
+ */
+public class SeekCmd extends MusicCommand
+{
+    public SeekCmd(Bot bot)
+    {
+        super(bot);
+        this.name = "seek";
+        this.help = "seeks the current song" +
+                "\nExamples: `1:02:23` `+1:10` `-90`";
```
jagrosh: I wouldn't include this second line in the help, as every other command has a single-line brief help. Save the examples for if someone enters the command wrong (after the 'Expected format' error message later in this file)

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
jagrosh: This message is fine, it should match the RemoveCmd

src/main/java/com/jagrosh/jmusicbot/commands/music/SeekCmd.java
```diff
@@ -0,0 +1,81 @@
...
+        {
+            event.replyError("You cannot seek **" + playingTrack.getInfo().title + "** because you didn't add it!");
+            return;
+        }
+
+        String args = event.getArgs();
+        TimeUtil.SeekTime seekTime = TimeUtil.parseTime(args);
+        if (seekTime == null)
+        {
+            event.replyError("Invalid seek! Expected format: " + arguments);
```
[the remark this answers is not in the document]
jagrosh: Put the second line of the help here instead of in the help itself.

src/test/java/com/jagrosh/jmusicbot/TimeUtilTest.java
```diff
@@ -0,0 +1,49 @@
+package com.jagrosh.jmusicbot;
```
jagrosh: license header

src/test/java/com/jagrosh/jmusicbot/TimeUtilTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    public void relativeSeekingBackward() {
+        TimeUtil.SeekTime seek = TimeUtil.parseTime("-5");
+        assertNotNull(seek);
+        assertTrue(seek.relative);
+        assertEquals(-5000, seek.milliseconds);
+    }
+}
```
jagrosh: Good tests so far, although I also want to see `assertFalse(seek.relative)` for the seeks that are not relative, as well as some things that intentionally return null from parseTime

Timardo:
So, is this finally able to become a real feature? I am really looking forward to this functionality.

realmar:
Really looking forward to this feature! I tested it and get the following exception when trying to play a song, eg. `!!play hello`:

```
[18:14:39] [ERROR] [LocalAudioTrackExecutor]: Error in playback of YQHsXMglC9A
com.sedmelluq.discord.lavaplayer.tools.FriendlyException: Something broke when playing the track.
 at com.sedmelluq.discord.lavaplayer.tools.ExceptionTools.wrapUnfriendlyExceptions(ExceptionTools.java:43)
 at com.sedmelluq.discord.lavaplayer.track.playback.LocalAudioTrackExecutor.execute(LocalAudioTrackExecutor.java:117)discord_music_bot_1  |  at com.sedmelluq.discord.lavaplayer.player.DefaultAudioPlayerManager.lambda$executeTrack$1(DefaultAudioPlayerManager.java:349)
 at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
 at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
 at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.NullPointerException: null
 at java.base/java.net.URI$Parser.parse(URI.java:3104)
 at java.base/java.net.URI.<init>(URI.java:600)
 at com.sedmelluq.discord.lavaplayer.source.youtube.YoutubeTrackFormat.getUrl(YoutubeTrackFormat.java:66)
 at com.sedmelluq.discord.lavaplayer.source.youtube.YoutubeSignatureCipherManager.resolveFormatUrl(YoutubeSignatureCipherManager.java:97)
 at com.sedmelluq.discord.lavaplayer.source.youtube.YoutubeAudioTrack.loadBestFormatWithUrl(YoutubeAudioTrack.java:87)
 at com.sedmelluq.discord.lavaplayer.source.youtube.YoutubeAudioTrack.process(YoutubeAudioTrack.java:42)
 at com.sedmelluq.discord.lavaplayer.track.playback.LocalAudioTrackExecutor.execute(LocalAudioTrackExecutor.java:105)discord_music_bot_1  |  ... 4 common frames omitted
```

Whew-Inc:
@realmar My fork wasn't up to date with the latest changes. It should work now.

As for the PR, I think it can be considered abandoned, unless @jagrosh approves it or someone else picks it up (feel free to do so!). I have no interest in the project anymore since I've quit Discord.

realmar:
@Whew-Inc  Fantastic, it works! Thanks a lot, I'm very happy with your work :D

Khaled-20:
Is this still open or has it been implemented yet? I would love to see this feature and would be happy to help with it as needed.

jagrosh:
> Is this still open or has it been implemented yet?  

The PR is still open because it isn't finished yet, as per [this comment](https://github.com/jagrosh/MusicBot/pull/344#pullrequestreview-346667866)

> This is a feature I have kept on the backburner because just adding this command is not quite enough. The following are all related to seeking and should all be added at the same time:
> 
>     * [x]  seeking to a specific point in a song (what this command currently does)
> 
>     * [x]  seeking forward/backward a certain amount of time in a song
> 
>     * [ ]  starting a song at a specific time (both via the `play` command as well as in playlists; youtube has a format for this and jmusicbot should probably adapt that same format)
> 
> 
> Additionally, I want to keep time-related notation consistent throughout the bot.

ribbanya:
Is there a quick workaround to seek within the current track using `eval`?

MichailiK:
> EDIT: Okay I figured something out, for anyone else who stumbles onto this...

Seems too hacky for me.

> As for the PR, **I think it can be considered abandoned**, unless jagrosh approves it or someone else picks it up (feel free to do so!). **I have no interest in the project anymore since I've quit Discord**.

Alright, I'm picking it back up in #344.

Sanduhr32:
> ```js
> !eval // ```js
> var jump = {
>   relative: true,
>   hours: 0,
>   minutes: 0,
>   seconds: 0,
>   milliseconds: 0
> };
> // Seek code
> var track=guild.getAudioManager().getSendingHandler().getPlayer().getPlayingTrack();function jumpToMs(e){return 36e5*e.hours+6e4*e.minutes+1e3*e.seconds+e.milliseconds}function jumpToStr(e,t){var r=jumpToMs(t),a=t.relative?r<0?"-":"+":"→";return msToString(e)+" "+a+" "+msToString(r)}function pad(e,t,r){return r=r||"0",(e+="").length>=t?e:new Array(t-e.length+1).join(r)+e}function msToString(e){e=Math.abs(e);var t=Math.floor(e/36e5);e-=36e5*t;var r=Math.floor(e/6e4);e-=6e4*r;var a=Math.floor(e/1e3);e-=1e3*a;var n="";return t>0&&(n+=pad(t,2)+":"),r>0&&(n+=pad(r,2)+":"),n+pad(a,2)+"."+pad(e,3)}var status="Can't seek";if(track.isSeekable()){pos=track.getPosition();var newPos=36e5*jump.hours+6e4*jump.minutes+1e3*jump.seconds+jump.milliseconds;jump.relative&&(newPos+=pos),track.setPosition(Math.max(0,newPos)),status=jumpToStr(pos,jump)}
> //```

Maybe don't inline everything? Also a native implementation is better than using eval for this.

Also Java isn't JS. It would have been better if you picked the PR up instead of designing a roundabout for a straight path.

MichailiK:
but what if the scripting engine is js? :o

ribbanya:
> Maybe don't inline everything? Also a native implementation is better than using eval for this.
> 
> Also Java isn't JS. It would have been better if you picked the PR up instead of designing a roundabout for a straight path.

I'm not a contributor, just a user. I listen to long-form audio so I needed a way to seek without restarting the whole track. I only posted it on the off chance someone would have a similar use case while the pull request is ongoing.

antaxiom:
Hello, I noticed the last there hasn't been any comments on this issue since April, any updates?

MichailiK:
I will work on this feature in #674 in the future.

MichailiK:
Since my PR #674 is (mostly) finished, I'll close this in favor of it.

PR #359, opened 2020-03-07
Add single repeat

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
This pull request adds a new repeat mode: one/single repeat mode, which, when the track ends, re-adds the current track to the beginning of the queue.

I replaced the boolean repeat with an enum "RepeatMode", which consists of the fields `On`, `Off` and `One`. This breaks compatibility with server settings from previous versions. I have implemented [a (sloppy?) fix](https://github.com/jagrosh/MusicBot/pull/359/commits/d3bb3808a87039210b9d8530ef38187342767829#diff-302f56ced8e4594e533143ecc0de3c27R45-R51) for that.

### Purpose
Uhh... In the event of loving one song so much you would like to listen to it several times, I guess?

### Relevant Issue(s)
This PR has no relevant issues. It is listed at the [JMusicBot Features](https://github.com/jagrosh/MusicBot/projects/1) project.


src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
```diff
@@ -40,14 +40,24 @@ public SettingsManager()
...
                 JSONObject o = loadedSettings.getJSONObject(id);
+
+                // Legacy version support: On versions 0.2.7 and older, the repeat mode was represented as a boolean.
+                try
+                {
+                    if(o.getBoolean("repeat"))
+                        o.put("repeat", RepeatMode.On);
+                    else
+                        o.put("repeat", RepeatMode.Off);
+                } catch(JSONException e) { /* ignored */ }
```
MichailiK: I'm not feeling comfortable about this bit. I can't really think of another solution without storing a "settings version" on the server settings.

src/main/java/com/jagrosh/jmusicbot/settings/RepeatMode.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ *
+ * @author Michaili K
+ */
+public enum RepeatMode
+{
+    Off,
+    On,
+    One
```
MichailiK: Shall I make the fields uppercase or keep it like that?

src/main/java/com/jagrosh/jmusicbot/commands/general/SettingsCmd.java
```diff
@@ -59,7 +59,7 @@ protected void execute(CommandEvent event)
                         + "\nVoice Channel: " + (vchan == null ? "Any" : "**" + vchan.getName() + "**")
                         + "\nDJ Role: " + (role == null ? "None" : "**" + role.getName() + "**")
                         + "\nCustom Prefix: " + (s.getPrefix() == null ? "None" : "`" + s.getPrefix() + "`")
-                        + "\nRepeat Mode: **" + (s.getRepeatMode() ? "On" : "Off") + "**"
+                        + "\nRepeat Mode: **" + s.getRepeatMode() + "**"
```
MichailiK: Shall we make it bold when the repeat mode is not off instead?

src/main/java/com/jagrosh/jmusicbot/audio/AudioHandler.java
```diff
@@ -153,10 +154,14 @@ public boolean playFromDefault()
...
+        RepeatMode repeatMode = manager.getBot().getSettingsManager().getSettings(guildId).getRepeatMode();
         // if the track ended normally, and we're in repeat mode, re-add it to the queue
-        if(endReason==AudioTrackEndReason.FINISHED && manager.getBot().getSettingsManager().getSettings(guildId).getRepeatMode())
+        if(endReason==AudioTrackEndReason.FINISHED && repeatMode != RepeatMode.Off)
         {
-            queue.add(new QueuedTrack(track.makeClone(), track.getUserData(Long.class)==null ? 0L : track.getUserData(Long.class)));
+            if(repeatMode == RepeatMode.On)
+                queue.add(new QueuedTrack(track.makeClone(), track.getUserData(Long.class)==null ? 0L : track.getUserData(Long.class)));
+            else
+                queue.addAt(0, new QueuedTrack(track.makeClone(), track.getUserData(Long.class)==null ? 0L : track.getUserData(Long.class)));
```
jagrosh, line 164:
```suggestion
            QueuedTrack clone = new QueuedTrack(track.makeClone(), track.getUserData(Long.class)==null ? 0L : track.getUserData(Long.class));
            if(repeatMode == RepeatMode.On)
                queue.add(clone);
            else
                queue.addAt(0, clone);
```

src/main/java/com/jagrosh/jmusicbot/commands/general/SettingsCmd.java
jagrosh: Yeah

src/main/java/com/jagrosh/jmusicbot/settings/RepeatMode.java
jagrosh: They should be uppercase. Let's also add the emojis here, something along the lines of
```java
public enum RepeatMode
{
    OFF(""), ON("\uD83D\uDD01"), ONE("\uD83D\uDD02"); // 🔁, 🔂

    private final String emoji; 

    private RepeatMode(String emoji)
    {
        this.emoji= emoji;
    }

    public String getEmoji()
    {
        return emoji;
    }
}
```

src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
jagrosh: This seems alright. Another way to handle it might be to rename the repeat mode field from `repeat` to `repeat_mode` so that if `repeat` exists, we can always expect it to be a boolean.

jagrosh (changes requested):

MichailiK:
Due to unresolvable dependencies in dead maven repositories, as well as no longer having these dependencies in my cache, it has become increasingly difficult to contribute to MusicBot. I may open this pull request again, once the dead repositories have been replaced & all dependencies can be downloaded again.

src/main/java/com/jagrosh/jmusicbot/settings/RepeatMode.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ *
+ * @author Michaili K
+ */
+public enum RepeatMode
+{
+    OFF(null, "Off"),
+    ON("\uD83D\uDD01", "On"), // 🔁
+    ONE("\uD83D\uDD02", "One"); // 🔂
```
MichailiK, line 26: Should we call it "single" instead? "on" and "one" are very similar and could be confused.

jagrosh: OFF / ALL / SINGLE would probably work best

MichailiK:
Merging under behalf of @jagrosh

Issue #242, opened 2019-06-28
[Feature Request] Search Files In Local Folder


YuutoTheSage:
Similar to searching links in Youtube or SoundCloud, but for local files.

The command would be something like:
?localsearch <query>

Options that matches the query will pop up like when using the ?search command, then the user gets to choose which song that will get added to the queue.

If there is only one match to the query, the song will be added to the queue without confirmation.
If there are no matches to the query, the bot will say "No results found."

WhoPlaysTheFool:
bumping this due to the current situation with YT not working

PR #404, opened 2020-04-29
Add localsearch command

blackbracken:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR adds a command `?localsearch` which behaves `?search` for local files (ref. #242).

### Purpose
Allows to search from local files.

### Relevant Issue(s)
This PR closes issue #242 .

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -109,6 +109,7 @@ public class JMusicBot
                         new RemoveCmd(bot),
                         new SearchCmd(bot),
                         new SCSearchCmd(bot),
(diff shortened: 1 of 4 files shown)


jagrosh:
I'm a bit hesitant to allow anyone using the bot effectively walk the filesystem of the host looking for keywords. If there is still interest in this feature and PR, I think there should be a number of checks in place:
1. I think that this would need to be explicitly enabled in the config file
2. I think the starting path would also need to be explicitly set in the config file, or default to something like `./Songs` to avoid walking where we shouldn't
3. I think a list of valid file extensions should be provided somewhere (could also be set in the config) so that non-audio files are ignored.

blackbracken:
Thank you for your response. It might be a little while but I will implement according to the checks.

PR #439, opened 2020-06-09
Refactor BotConfig

dotneutron:
### This pull request...
  - [ ] Fixes a bug
  - [x] Boosts code quality or performance

### Description
I noticed quite a few hard-coded strings, so I decided to propose a very common solution, known as localization or internationalization (_i18n_). It can be used to further declutter the rest of the codebase. That lead me to other incremental changes until I got to this form. I wouldn't call it final, I just ran out of ideas. 😀

There was some stuff along the way that I didn't really consider to be essential (like `START_TOKEN` and `END_TOKEN`), so I just removed it. Feel free to correct me if I'm wrong.

There might have been some minor oversight in the past, like trailing whitespace on otherwise empty lines and some other miscellaneous details. I'm currently using IntelliJ IDEA + SonarLint and I can say it's quite good at catching any developer's hiccups.

I tried to limit myself to a single class, to make the review hassle-free.

### Purpose
Reduce cognitive complexity of `BotConfig#load()` and improve overall readability.

### Relevant Issue(s)
**null**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,11 @@
             <version>1.3.2</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.10</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.12</version>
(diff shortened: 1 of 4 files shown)


jagrosh:
I'm going to close this PR, not because these changes are unwanted, but rather because a lot of the things in it are already planned to be implemented at other levels. Internationalization is a planned feature for JDA-Utilities, for which I'm already building a wrapper around resource bundles (and when that is completed, I'll be refactoring the entire project to make use of it). START_TOKEN and END_TOKEN are required, as the autogenerated reference config is a concatenation of the configs from all dependencies; these tokens are used to isolate just JMusicBot's config for creating the front-facing config.txt for the user. Finally, I'm not a big fan of introducing extra dependencies (ie commons-lang3) just to simplify one line of code.

PR #292, opened 2019-10-15
Fix bot joining in AFK voice channels

MichailiK:
### This pull request...
  - [x] Fixes a bug
  - [ ] Introduces a new feature
  - [ ] Improves an existing feature
  - [ ] Boosts code quality or performance

### Description
Since Discord automatically mutes users in AFK channels, bots shouldn't
play music inside AFK channels. This PR also fixes #147.
**This may be a controversial PR, since it takes away the freedom of using music commands in AFK channels.**

### Purpose
This prevents users from playing music in AFK channels, where all users are guild-muted.

### Relevant Issue(s)
This PR closes #147.

PR #470, opened 2020-07-25
Revert "Fix bot joining in AFK voice channels"

Aur3m:
Reverts jagrosh/MusicBot#292

Pls I can't use the bot in an afk channel with the loop mode after having unmute him


Sanduhr32:
This PR in it's current state seems invalid.

jagrosh:
It is intentional that the bot does not join afk channels.

PR #538, opened 2020-10-30
Next song feature

chr-ibb:
### This pull request...
  - [ ] Fixes a bug
  - [X ] Introduces a new feature

### Description
Adds a new dj command, "next," which skips the currently playing track and adds it to the end of the queue if in repeat mode. 

NextCmd is mostly based off ForceskipCmd, but it also calls handler.addTrackIfRepeat(track), which is a new method in AudioHandler. AudioHandler's onTrackEnd method now uses addTrackIfRepeat appropriately. addTrackIfRepeat uses some of the same logic that was previously in onTrackEnd, that's why I consolidated. 

### Purpose
When listening to a playlist on repeat, it is often desirable to skip a song while keeping it in the repeating queue. Otherwise, every time you skip a song, the repeating playlist gets smaller and smaller. Both features are useful, so the "next" dj command is purely an addition. 

### Relevant Issue(s)
None that I found

###  Other Comments
This is my first ever pull request, please let me know if I've made any mistakes. 
I did not make any changes to the LICENSE file or ConsolePanel.java, I'm not sure why it says so in the diffs (its just the same file removed and added). Thanks for the great music bot.


chr-ibb:
I did some basic tests, just using !next a bunch with repeat on and off, and with different sized queues. Playing music from  youtube has been buggy lately, so all of my tests were done with local files.

jagrosh (changes requested):
A few issues with this PR:
- There are several files (including the license) that this PR seems to have rewritten without changing anything (maybe a line-endings issue?)
- I don't know if `next` is quite right for the name of this command; I'm open to discussion for a better name for it
- `NextCmd.java` needs to follow the same conventions as the rest of the project (license header, @ author tag, { on new line, etc)

Overall I think this command is an interesting idea, but there's a number of preliminary changes/cleanup that would need to be done before I can delve into the actual implementation presented here.

chr-ibb:
Thank you for the response! I've rebased my feature branch and corrected some things based on your feedback.

- Fixed the issue with 3 unchanged files showing as changed in diffs. Your intuition was correct- those 3 exact files have CRLF line endings in your index, while my index just had LF line endings across the board. If you don't mind me asking- is that intentional? (Genuine question, I'm a novice). 
- I'm 100% flexible on the command name, and I have some suggestions below. 
- I updated NextCmd.java to follow the conventions of the project.

### Command name:
I originally chose the name "next" because in the media players I'm familiar with, pressing the forward arrow button (often labeled "next") while repeat/loop is on will skip the song without removing it from the collection of songs being looped. However, I realize that we aren't dealing with a standard music player. In standard media players that functionality is present when playing from a preset collection of tracks, like an album or playlist, but in this bot everything is added to a queue. 

My suggestions for alternative names:
"advance"
"continue"
"proceed"
"progress"

I'll see if I can come up with something better, but also I'm completely open to a name of your choosing.

chr-ibb:
Looking further, I'm curious if you'd be interested in having three different commands with this functionality: a Music Command that mirrors "SkipCmd," a DJ Command that mirrors "ForceskipCmd," and another DJ Command that mirrors "SkiptoCmd." I'd be happy to include them in this PR or a new one.

chr-ibb:
Another potential solution would be to forgo adding this new functionality to a new command, and instead change the existing "skip" commands to work this way (re-add skipped songs to end of queue if repeat is ON). Then, you could allow the existing "remove" commands to accept no argument or an argument of 0, which would remove the playing song _and_ skip it. This solution is attractive because it would also allow for using commands like ForceRemoveCmd to remove the song that is currently playing (right now, if user#1234 adds 10 songs to an empty queue, and then an admin/DJ uses "!forceremove user#1234", it will only remove 9 songs because the first song is already playing). If making this change to the "remove" commands is not desirable, an alternative would be to add a new command for skipping _and_ removing. One issue with that approach is the new command would be redundant while repeat is OFF.  

Pros of this solutions would be having a "skip" that works more similarly to media players when repeat is ON, while not losing the ability to skip _and_ remove if that is desired. While repeat is OFF, "skip" will still behave the same way that it does currently.

Cons of this solution would be potentially confusing users who are used to the current functionality of "skip" and "remove", or  having a command that is redundant while repeat is OFF (if a new command is added).

PR #572, opened 2020-12-09
fix "[ERROR] [LocalAudioTrackExecutor]: Error in playback"

kriegmaster56:
change lavaplayer version from 1.3.55 to 1.3.63

This pull requestFixes a bug

### Purpose
There's a probleme in version 0.3.2 JMusicBot where's it won't load certain song from youtube, mostly "(Official Video)" song.

the upgrade in version for lavaplayer fix the problem

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <dependency>
             <groupId>com.sedmelluq</groupId>
             <artifactId>lavaplayer</artifactId>
-            <version>1.3.55</version>
+            <version>1.3.63</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


MichailiK (approved):
This should also fix a few other issues, like playing age-restricted videos & loading playlists with more than 100 entries.

Sanduhr32:
@jagrosh might wanna merge this. (And why did you stop making OS specific jars with reduced binaries?)

jagrosh:
This is already updated in my local branch, so the next update will use the latest lavaplayer

PR #579, opened 2021-01-06
Add Playlist appendq command

jhleekr:
Added command 'playlist appendq <name> - appends songs in queue to an existing playlist'

### This pull request...
  - [ ] Fixes a bug

### Description
Added command 'playlist appendq <name> - appends songs in queue to an existing playlist'
Command playlist appendq <name> will append playing song and songs in queue if exist to an existing playlist.

### Purpose
It makes the user can save the current queue as a playlist so that it can be recalled and used later.

### Relevant Issue(s)
None

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
@@ -194,7 +194,7 @@ public class PlaylistCmd extends OwnerCommand
             String pname = event.getArgs().replaceAll("\\s+", "_");
             Playlist playlist = bot.getPlaylistLoader().getPlaylist(pname);
             if(playlist==null)
-                event.replyError(event.getClient().getError()+" Playlist `"+pname+"` doesn't exist!");
+                event.replyError("Playlist `"+pname+"` doesn't exist!");
             else
             {
                 AudioHandler ah = (AudioHandler)event.getGuild().getAudioManager().getSendingHandler();
@@ -204,7 +204,7 @@ public class PlaylistCmd extends OwnerCommand
                     Message nowp = ah.getNowPlaying(event.getJDA());
                     if(nowp==null)
                     {
-                        event.replyWarning(event.getClient().getWarning() + " There is no music in the queue!");
+                        event.replyWarning("There is no music in the queue!");
                         return;
                     }
                 }
@@ -228,11 +228,11 @@ public class PlaylistCmd extends OwnerCommand
                 try
                 {
                     bot.getPlaylistLoader().writePlaylist(pname, builder.toString());
-                    event.replySuccess(event.getClient().getSuccess()+" Successfully added "+al.size()+" items to playlist `"+pname+"`!");
+                    event.replySuccess("Successfully added "+al.size()+" items to playlist `"+pname+"`!");
                 }
                 catch(IOException e)
                 {
-                    event.replyError(event.getClient().getError()+" I was unable to append to the playlist: "+e.getLocalizedMessage());
+                    event.replyError("I was unable to append to the playlist: "+e.getLocalizedMessage());
                 }
             }
         }
--- a/src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
@@ -41,7 +41,7 @@ public class PlaylistCmd extends OwnerCommand
         this.bot = bot;
         this.guildOnly = false;
         this.name = "playlist";
-        this.arguments = "<append|appendq|delete|make|setdefault>";
+        this.arguments = "<append|appendqueue|delete|make|setdefault>";
         this.help = "playlist management";
         this.aliases = bot.getConfig().getAliases(this.name);
         this.children = new OwnerCommand[]{
@@ -194,7 +194,7 @@ public class PlaylistCmd extends OwnerCommand
             String pname = event.getArgs().replaceAll("\\s+", "_");
             Playlist playlist = bot.getPlaylistLoader().getPlaylist(pname);
             if(playlist==null)
-                event.reply(event.getClient().getError()+" Playlist `"+pname+"` doesn't exist!");
+                event.replyError(event.getClient().getError()+" Playlist `"+pname+"` doesn't exist!");
             else
             {
                 AudioHandler ah = (AudioHandler)event.getGuild().getAudioManager().getSendingHandler();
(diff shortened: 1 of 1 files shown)


phamleduy04:
I got some IP bans in my VPS because my queue has about 100 songs in it. So I wouldn't recommend this command.

jhleekr:
I'm not sure what it means. This command will just export the songs in the queue to a playlist text file so user can load them the same later.

phamleduy04:
I have a playlist with like 100 songs, and it will save 100 songs in the playlist. When I load it, the bot will send 100 requests to youtube and that's made youtube returns some 429 code.

jhleekr:
Oh, i see. I think the problem you're talking about is the playlist function itself rather than this command. It certainly needs improvement.

jhleekr:
Hi @jagrosh, this is a simple commit, can you review it?

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,75 @@ protected void execute(CommandEvent event)
...
+        @Override
+        protected void execute(CommandEvent event) 
+        {
+            String[] parts = event.getArgs().split("\\s+", 2);
+            if(parts.length<1)
+            {
+                event.reply(event.getClient().getError()+" Please include a playlist name to add!");
+                return;
+            }
+            String pname = parts[0];
```
MichailiK, line 200: For consistency with the other playlist commands that also only take a list name as an argument (`make`, `delete` etc.), replace all spaces with an underscore.

image

```suggestion
            String pname = event.getArgs().replaceAll("\\s+", "_")
```

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,75 @@ protected void execute(CommandEvent event)
...
+            Playlist playlist = bot.getPlaylistLoader().getPlaylist(pname);
+            if(playlist==null)
+                event.reply(event.getClient().getError()+" Playlist `"+pname+"` doesn't exist!");
```
MichailiK, line 203: Some old code may still use this, but it is preferred to use `event.replySuccess`, `event.replyWarning` or `event.replyError` now.

```java
event.replyError("Playlist `"+pname+"` doesn't exist!");
```

Replace all `event.reply`s across the PR.

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,75 @@ protected void execute(CommandEvent event)
...
+                playlist.getItems().forEach(item -> builder.append("\r\n").append(item));
+                for(AudioTrack url: al)
+                {
+                    if(url==null)
+                    {
+                        continue;
+                    }
+                    String u = url.getInfo().uri;
+                    if(u.startsWith("<") && u.endsWith(">"))
+                        u = u.substring(1, u.length()-1);
```
MichailiK, line 233: This is not needed. The angle brackets are already removed by the play command:

https://github.com/jagrosh/MusicBot/blob/d5bb16d1436c9e5dcd81df8221d38cb7fa4317c7/src/main/java/com/jagrosh/jmusicbot/commands/music/PlayCmd.java#L87-L89

```suggestion

```

MichailiK (changes requested):

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,67 @@ protected void execute(CommandEvent event)
...
+            this.guildOnly = false;
+        }
+
+        @Override
+        protected void execute(CommandEvent event) 
+        {
+            String pname = event.getArgs().replaceAll("\\s+", "_");
+            Playlist playlist = bot.getPlaylistLoader().getPlaylist(pname);
+            if(playlist==null)
+                event.replyError(event.getClient().getError()+" Playlist `"+pname+"` doesn't exist!");
```
MichailiK, line 197: Whoops! This will result in a message like:
> ⚠️⚠️ Playlist `name` doesn't exist!

`event.replyError("text")` adds the emoji to the beginning for you, which is much better than to do `event.reply(event.getClient().getError()+" text")`. This function didnt exist a few years ago, some old code still uses that. 

Nowadays, you just wanna do `event.replyError("text")`.

```suggestion
                event.replyError("Playlist `"+pname+"` doesn't exist!");
```

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,67 @@ protected void execute(CommandEvent event)
...
+            else
+            {
+                AudioHandler ah = (AudioHandler)event.getGuild().getAudioManager().getSendingHandler();
+                List<QueuedTrack> list = ah.getQueue().getList();
+                if(list.isEmpty())
+                {
+                    Message nowp = ah.getNowPlaying(event.getJDA());
+                    if(nowp==null)
+                    {
+                        event.replyWarning(event.getClient().getWarning() + " There is no music in the queue!");
```
MichailiK, line 207: Same mistake here
```suggestion
                        event.replyWarning("There is no music in the queue!");
```

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,67 @@ protected void execute(CommandEvent event)
...
+                    {
+                        continue;
+                    }
+                    String u = url.getInfo().uri;
+                    builder.append("\r\n").append(u);
+                }
+                try
+                {
+                    bot.getPlaylistLoader().writePlaylist(pname, builder.toString());
+                    event.replySuccess(event.getClient().getSuccess()+" Successfully added "+al.size()+" items to playlist `"+pname+"`!");
```
MichailiK, line 231: ... and here
```suggestion
                    event.replySuccess("Successfully added "+al.size()+" items to playlist `"+pname+"`!");
```

src/main/java/com/jagrosh/jmusicbot/commands/owner/PlaylistCmd.java
```diff
@@ -169,6 +176,67 @@ protected void execute(CommandEvent event)
...
+                }
+                catch(IOException e)
+                {
+                    event.replyError(event.getClient().getError()+" I was unable to append to the playlist: "+e.getLocalizedMessage());
```
MichailiK, line 235: ... and finally here
```suggestion
                    event.replyError("I was unable to append to the playlist: "+e.getLocalizedMessage());
```

MichailiK (changes requested):
Almost! It seems like I didnt bring the message well across with `event.replySuccess`, `event.replyWarning` etc.. There's a mistake with the replies that need to be fixed.

MichailiK (approved):
Looks good! @jagrosh will have to take a look at it & will merge if it's all good.

PR #582, opened 2021-01-12
Allow lyrics to be available anytime anywhere, fix typos

giovanniwijaya:
### This pull request...
  - [ ] Fixes a bug

### Description
In the present code, lyrics are only available when something is playing in the guild. I modified placement of `guildOnly` property from the MusicCommands.java to the individual files in commands/music folder, and set `bePlaying` to false in LyricsCmd.java

In addition, I fixed some typos in the `help` command (such as "Soundcloud" to "SoundCloud", "bots settings" to "bot’s settings", etc)

### Purpose
Sometimes users want to look up lyrics while listening to the song elsewhere, so it's better to not limit the `lyrics` functionality.

### Relevant Issue(s)
None, just an improvement

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
@@ -38,7 +38,6 @@ public abstract class MusicCommand extends Command
     public MusicCommand(Bot bot)
     {
         this.bot = bot;
-        this.guildOnly = true;
         this.category = new Category("Music");
     }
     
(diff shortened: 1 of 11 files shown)


giovanniwijaya:
Turns out there's an issue with sending embeds to DMs, but sending just text works. I don't know what the issue is, as no error messages are output and the docs don't suggest that this issue may arise, but I also don't have any workarounds, so for now I've split the workings of LyricsCmd.java into two: if in guild send embeds (max 2048 characters length each), if in DM send texts (max 2000 characters length each).

Also when using embeds, I've added code to show the lyrics source at the footer of the very last embed sent.

jagrosh:
This PR should not be modifying how MusicCommand works. If the goal is for lyrics to be available at all times, then it shouldn't inherit from MusicCommand.

jagrosh (changes requested):
See my earlier comment; this feature should not modify how MusicCommand works.

MichailiK:
Since #940 has been merged, this pull request is now obsolete.

Issue #585, opened 2021-01-15
[Feature Request] ZPL playlist Support


LAKSHAY184:
**Is your feature request related to a problem? Please describe.**
I want to use a groove music playlist on a discord bot.
**Describe the solution you'd like**
Since, Groove stores the playlists in zpl files, it would be easy to just upload the zpl file in the folder and get it running.

**Describe the basic flow/steps of using this feature**
How this feature would be used:
1. this could be used to run groove music playlists
2. 
3. 

**Additional context**

**Have you confirmed that this idea is not already described here?**
https://github.com/jagrosh/MusicBot/projects/1
This feature is not here

pbanj:
You can probably just change the file extension

LAKSHAY184:
No, It can't be done because it is written in a zpl format which looks a bit similar to xml

pbanj:
Ah, https://gist.github.com/jgarza9788/a17de65f41f14662491e this might be helpful for the time being.

LAKSHAY184:
Thanks!

LAKSHAY184:
> Ah, https://gist.github.com/jgarza9788/a17de65f41f14662491e this might be helpful for the time being.

hi! I was after some hard work able to change my playlist to m3u but it isn't working with the music bot
Can you guide me!

PR #649, opened 2021-03-27
Added a ZPL to Txt convertor for zune Playlists

lakshaycodes:
### This pull request...

### Description
This Pull Request helps in handling Groove Music Playlists.

### Purpose
Using the Python Script that I added, users can import groove music Playlists to the Music Bot just by changing the extension of the zpl file generated by groove music to txt. Then, the python file will red the file and will take out the paths of the songs from the python file and store them into a new file that is supported by the bot.

### Relevant Issue(s)
This PR closes issue #585

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,0 +1,17 @@
+file1 = input("Give the full path to the zpl file")
+file2 = input("Give the name for the new text file")
+file = open(file1, 'r')
+x = file.readlines()
+file2 = open(file2, 'w')
+file2.write('')
+file2.close()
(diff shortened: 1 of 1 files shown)


jagrosh:
I don't think a python script is the best solution to this, for a few reasons:
- not every system has python by default
- most users will not know how to run this
- this will need to be downloaded separately from the release jar
- this will need to be edited each time someone wants to add a new playlist

The ideal solution to this problem (and other playlist formats) would be to modify the playlist loader to treat multiple formats as playlists and parse them accordingly.

PR #650, opened 2021-03-29
Add basic Dockerfile

Prof-Bloodstone:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR adds a very basic Dockerfile. Packaged it for myself, but someone else might also find it useful. 
It uses multi-stage builds, greatly decreasing the size of the final image. If you also want to provide docker image downloads, both docker-hub and quay.io has automatic builders so there isn't need to configure much.
I've decided to put the jar in the rootfs because I never really know of a better UNIXy place. Open to suggestions.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+!pom.xml
+!src
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+ARG JAVA_VERSION=11
+
+FROM maven:3.6-openjdk-$JAVA_VERSION as builder
+
+WORKDIR /tmp/workdir
+
+COPY pom.xml ./
+
+RUN mvn dependency:go-offline
+
+COPY src/ ./src/
+
+RUN mvn package
+
+
+ARG JAVA_VERSION
+
+FROM openjdk:$JAVA_VERSION-jre
+
+COPY --from=builder /tmp/workdir/target/JMusicBot-*-All.jar /JMusicBot.jar
+
+CMD ["java", "-Dnogui=true", "-jar", "/JMusicBot.jar"]


Sanduhr32 (changes requested):
1) minimum required java version is 8 (also openjdk 11 has ssl issues on some systems)
2) this builds src, which is kinda unsupported
3) probably doesn't need to build src each time

Prof-Bloodstone:
1. I chose Java 11 because it's the latest LTS release. There's no worry about it not working on some systems because, well, containers. Unless I just didn't find the case where it breaks yet. Java 11 has plenty of runtime improvements like GC, although I don't think Music Bot creates enough allocations to worry about it.
2. I'd rather avoid copying the compiled jar from the host - this limits amount of uses for the Dockerfile itself.
3. I tried to cache dependencies etc. I don't have any ideas on further limiting when the source is rebuilt.
4. Thanks a lot! I also noticed the bot creates more data than I expected, so in my deployment I've moved everything.

Also probably should switch to some random, non-root user in the Dockerfile.

Sanduhr32:
>     1. I chose Java 11 because it's the latest LTS release. There's no worry about it not working on some systems because, well, containers.
no its not the latest LTS release but whatever.

>     3. I tried to cache dependencies etc. I don't have any ideas on further limiting when the source is rebuilt.
maybe check if the jar exists in your workdir, if not run maven for building the jar, else skip to executing the jar.

4. has been edited

Prof-Bloodstone:
1. What is the latest LTS then? Afaik next one is Java 17 scheduled for middle of this year, but currently it's Java 11.

3. Sadly it's not possible to check if built jar exists. What I could do is get rid of the builder all together and only download the jar from GH releases. Downside is that it only works for releases (can't use snapshots) and needs to be updated with each release, since GH doesn't have "latest" endpoint.

4. Good idea. I'm thinking about putting the jar in /usr/... and having workspace/volume set to /data/, though might go with /var/jmusicbot/ dir.

Thanks for all the feedback!

Sanduhr32:
2. Yeah well docker can check if files and or dirs exist. Maybe learn some more docker.

fitzychan:
@Prof-Bloodstone Sorry if I came off combative.  My response was more directed at Sanduhr32's requests to fix the docker file when what you have is pretty standard.

PR #733, opened 2021-07-29
Removed broken and unused repository and dependency

diademiemi:
### This pull request...
  - [X] Fixes a bug

### Description
This removes dl.bintray.com, which no longer exists. It also removes lavaplayer-natives-extra, since this is not in the repository from Lavaplayer, and it compiles and the bot works without, so it seems to be unused. 

### Purpose
With this PR, this project can be compiled with `mvn clean package`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,11 +17,6 @@
             <name>bintray</name>
             <url>https://jcenter.bintray.com</url>
         </repository>
-        <repository>
-            <id>bintray-sedmelluq-com.sedmelluq</id>
-            <name>bintray</name>
-            <url>https://dl.bintray.com/sedmelluq/com.sedmelluq</url>
-        </repository>
     </repositories>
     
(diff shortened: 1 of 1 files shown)


jagrosh:
The natives package is not unused; it results in additional natives being built into the final jar which enables the bot to run on several by-default-unsupported platforms, such as ARM processors.

MichailiK:
Is there any other (available) repository with that dependency? I keep removing that repo & dependency while doing PRs for MusicBot, and I occasionally commit my modified pom unintentionally.

diademiemi:
It doesn't seem to be at https://m2.dv8tion.net/, where the other lavaplayer-natives packages reside. It seems like there is no repository which provides this (https://github.com/sedmelluq/lavaplayer/issues/622), and you need to manually get the file in order to use this.

Issue #846, opened 2021-09-17
[Feature Request] Leave Command


BluepanicDV:
It is simple, a command so that the bot can leave the music channel, I know that it can be configured so that after X time the bot leaves the channel, but it would be more like a second option if this leave command is added, allowing that the servers that have finished do not consume resources of the bot's servers

Example
_I think it's a bit obvious, but I'm going to put it anyway_

Bot prefix: **?**

Command in config archive: **Leave**

Using the command: **?Leave**

According to the configuration file, the bot would simply leave the channel or leave it and send a message to the channel where the _?Leave_ command was sent.

that simple, I hope to explain it well...

MichailiK:
... `?stop`?

Issue #845, opened 2021-09-17
[Feature Request] Clear Queue Command


DragonDuck24:
a command like "clear" that would clear the entire queue

It would be similar to rhythm's clear command.

"?clear"
the bot would say "queue cleared!"
and all songs would stop

probably a dj command

jagrosh:
This is what the command `stop` does

DragonDuck24:
Oh, ok
thanks for the speedy reply

PR #847, opened 2021-09-17
Add "clear" alias to stop command

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
Adds a new alias for the `stop` command, "clear".

### Purpose
Some people seem to be unaware of `stop`'s existence, so it is perhaps a good idea to add "clear" as a default alias for the stop command.

### Relevant Issue(s)
See also #845 and #846

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -169,7 +169,7 @@ aliases {
   playnext = []
   repeat = []
   skipto = [ jumpto ]
-  stop = []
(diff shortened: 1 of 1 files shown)


jagrosh:
I am concerned that people will attempt to use this to clear messages

MichailiK:
Right, some music bots do that. If I recall correctly, such a feature has been requested too. I suppose leave would be a better alias.

PR #858, opened 2021-09-18
Use MKDocs to generate documentation instead of Jekyll

Dougley:
This PR introduces a new documentation generator, MKDocs, to replace Jekyll.

**A preview is available at https://musicbot.pages.dev**    
Cloudflare pages is used as an example since MKDocs is configured in a way that requires an apex domain, it will work fine on your GitHub pages config with jmusicbot.com.

### What's different?

- Unified documentation from the `gh-pages` branch to `master`.
  - MKDocs will overwrite the `gh-pages` branch, and since Jekyll is gone there's no reason to put Markdown in its own branch.
- Removed all references to Imgur and use locally served files instead.
  - [Imgur is not a CDN](https://imgur.com/tos), and with Pages there's little reason to outsource images since it adds complexity.
- Added a [GitHub Action](https://github.com/features/actions) that will automatically redeploy documentation when the `master` branch is updated.
- `README.md` uses the same assets as the docs.
  - Again, Imgur is not a CDN.

### What's not done?
- [ ] Navigational layouts
- [ ] Admonitions
- [ ] Markdown headers
- [ ] Google Analytics?

### Special changes

Similar to your Jekyll config, MKDocs will:
- Inject Google AdSense Syndication and OpenGraph/Twitter embed headers (see `docs/overrides/main.html`)
- Use Fira Sans

thanks you made me do it 🙂


MichailiK (changes requested):
Didn't @jagrosh want for documentation to be on a different branch, to prevent clogging up the commit history?

Dougley:
I've done it this way because, as far as I can see, Actions only trigger when the default branch is updated. 🤠

Dougley:
Closing in favour of #860

Issue #833, opened 2021-09-16
 Soundcloud commands do not work #714 


Coolgam3r2152:
Describe the bug
Nothing from soundcloud is able to be played,

To Reproduce

    play any soundcloud link or do the search soundcloud command

Expected behavior
expected to play the requested song link/result

Debug Info
```
System Properties:
  java.version = 1.8.0_252
  java.vm.name = OpenJDK 64-Bit Server VM
  java.vm.specification.version = 1.8
  java.runtime.name = OpenJDK Runtime Environment
  java.runtime.version = 1.8.0_252-b09
  java.specification.version = 1.8
  os.arch = amd64
  os.name = Linux

JMusicBot Information:
  Version = 0.3.5
  Owner = HIDDEN
  Prefix = ?
  AltPrefix = null
  MaxSeconds = 0
  NPImages = false
  SongInStatus = false
  StayInChannel = false
  UseEval = false
  UpdateAlerts = true

Dependency Information:
  JDA Version = 4.3.0_312
  JDA-Utilities Version = 3.0.5
  Lavaplayer Version = 1.3.78

Runtime Information:
  Total Memory = 31
  Used Memory = 28

Discord Information:
  ID = 785570156553961543
  Guilds = 3
  Users = 4
```
Additional context
Youtube works fine but any soundcloud link or search result is unable to be played by the bot.

extend2to3meters:
if you dont mind compiling the bot yourself, heres my [workaround](https://github.com/extend2to3meters/MusicBot/commit/855adcc67ce912c2479abd006a83a9bba2930e30#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R37-R41) with https://github.com/sedmelluq/lavaplayer/pull/650 merged

hermitoff:
For me Soundcloud work...

MichailiK:
We now maintain our own fork of lavaplayer that contains the above pull request. This issue will close once the upcoming version (0.3.6) is released with the fix.

MichailiK:
Version 0.3.6 has addressed this issue & [is now available for download](https://github.com/jagrosh/MusicBot/releases/latest).

PR #859, opened 2021-09-18
Add "Known issues" section to Troubleshooting

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR adds a "Known issues" section to the Wiki, which addresses known and on-going problems, like loading from SoundCloud (as of writing).

### Purpose
Users may be more likely referring to the Troubleshooting page, and see that their problem is known & being fixed, instead of opening duplicate issues or asking in the support server.

### Relevant Issue(s)
None necessarily.

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -9,6 +9,10 @@ permalink: /troubleshooting
 1. TOC
 {:toc}
 
+## Known issues
+### ⚠ SoundCloud URLs result in a "Error loading track." message!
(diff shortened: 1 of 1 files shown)


jagrosh:
This is literally what the issues tab on github is for. Instead of putting any specific issues here (which we would have to update every time there's a new issue on github) let's just add a section on the troubleshooting page to check for known issues

MichailiK:
From a UX  standpoint, looking at a sea of issues & using the search bar feels a bit too inconvenient. Perhaps we should start pinning ongoing issues that make the bot kinda unusable (right now for example #833).

jagrosh:
Yes, that's what pinning is for

PR #932, opened 2021-09-28
Voice channel mention

yo-louk:
### This pull request...
  - [ ] Fixes a bug

### Description
I recently noticed voice channels can be mentioned just like text channels. A small change to the code can implement this feature.

image
The top example is the current way the bot would mention a channel, the bottom is what I wish it would look like.
This is my first pull request. I have not tested this change as I consider it to be too simple and don't expect it to fail.

### Purpose
Aesthetics and ease of use: clicking on the mentioned channel will automatically connect you to it.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
@@ -70,7 +70,7 @@ public abstract class MusicCommand extends Command
             GuildVoiceState userState = event.getMember().getVoiceState();
             if(!userState.inVoiceChannel() || userState.isDeafened() || (current!=null && !userState.getChannel().equals(current)))
             {
-                event.replyError("You must be listening in "+(current==null ? "a voice channel" : "<#"+current.getId()+">")+" to use that!");
+                event.replyError("You must be listening in "+(current==null ? "a voice channel" : current.getAsMention())+" to use that!");
                 return;
             }
 
@@ -89,7 +89,7 @@ public abstract class MusicCommand extends Command
                 }
                 catch(PermissionException ex) 
                 {
-                    event.reply(event.getClient().getError()+" I am unable to connect to <#"+userState.getChannel().getId()+">!");
+                    event.reply(event.getClient().getError()+" I am unable to connect to "+userState.getChannel().getAsMention()+"!");
                     return;
                 }
             }
--- a/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
@@ -70,7 +70,7 @@ public abstract class MusicCommand extends Command
             GuildVoiceState userState = event.getMember().getVoiceState();
             if(!userState.inVoiceChannel() || userState.isDeafened() || (current!=null && !userState.getChannel().equals(current)))
             {
-                event.replyError("You must be listening in "+(current==null ? "a voice channel" : "**"+current.getName()+"**")+" to use that!");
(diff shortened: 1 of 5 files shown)


src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
```diff
@@ -70,7 +70,7 @@ protected void execute(CommandEvent event)
             GuildVoiceState userState = event.getMember().getVoiceState();
             if(!userState.inVoiceChannel() || userState.isDeafened() || (current!=null && !userState.getChannel().equals(current)))
             {
-                event.replyError("You must be listening in "+(current==null ? "a voice channel" : "**"+current.getName()+"**")+" to use that!");
+                event.replyError("You must be listening in "+(current==null ? "a voice channel" : "<#"+current.getId()+">")+" to use that!");
```
MichailiK, line 73:
```suggestion
                event.replyError("You must be listening in "+(current==null ? "a voice channel" : current.getAsMention())+" to use that!");
```

src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
```diff
@@ -89,7 +89,7 @@ protected void execute(CommandEvent event)
                 }
                 catch(PermissionException ex) 
                 {
-                    event.reply(event.getClient().getError()+" I am unable to connect to **"+userState.getChannel().getName()+"**!");
+                    event.reply(event.getClient().getError()+" I am unable to connect to <#"+userState.getChannel().getId()+">!");
```
MichailiK, line 92:
```suggestion
                    event.reply(event.getClient().getError()+" I am unable to connect to "+userState.getChannel().getAsMention()+"!");
```

src/main/java/com/jagrosh/jmusicbot/commands/admin/SetvcCmd.java
```diff
@@ -62,7 +62,7 @@ else if (list.size()>1)
             else
             {
                 s.setVoiceChannel(list.get(0));
-                event.reply(event.getClient().getSuccess()+" Music can now only be played in **"+list.get(0).getName()+"**");
+                event.reply(event.getClient().getSuccess()+" Music can now only be played in <@"+list.get(0).getId()+">");
```
MichailiK, line 65:
```suggestion
                event.reply(event.getClient().getSuccess()+" Music can now only be played in "+list.get(0).getAsMention());
```

MichailiK (changes requested):
JDA offers a method to get a channel's mention string. Use that method instead.

MichailiK:
Additionally, my IDE has found another 3 usages that you've missed:

https://github.com/jagrosh/MusicBot/blob/b83028fb97318825c7adaa12caa4a3b94513c652/src/main/java/com/jagrosh/jmusicbot/audio/AudioHandler.java#L205

https://github.com/jagrosh/MusicBot/blob/b83028fb97318825c7adaa12caa4a3b94513c652/src/main/java/com/jagrosh/jmusicbot/commands/general/SettingsCmd.java#L60

https://github.com/jagrosh/MusicBot/blob/b83028fb97318825c7adaa12caa4a3b94513c652/src/main/java/com/jagrosh/jmusicbot/utils/FormatUtil.java#L77

Update these accordingly.

yo-louk:
Thanks for the elaborate feedback.
I agree with `getAsMention()` being more appropriate than `"<#"+vc.getId()+">"`.
I also applied these changes to the 3 other usages that I had missed earlier.

After successfully compiling the project, I ran the commands and got some good results, a few of which you can see below:

image

MichailiK (approved):
Looks good! @jagrosh will have to take another look & merge if everything looks good.

PR #940, opened 2021-10-01
Allow usage of LyricsCmd without playing music

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
Allows using the `lyrics` command without music playing.

### Purpose
Ability to get the lyrics of a song by using `lyrics [song name]`.

### Relevant Issue(s)
None


yo-louk:
This doesn't exactly have to do with your request, but figured I'd throw my suggestion in here anyway:

image
Is it possible to code the bot in such way that it remembers the song request "bad habits" and attempts to look up lyrics for that, because looking up the lyrics by its YouTube title doesn't always return the lyrics (as seen above). This can also be a secondary method, in case the YouTube title lookup method fails. This way the last command could be omitted.

MichailiK:
That would be somewhat complex/bothersome & could perhaps cause confusion. I believe the current behavior is fine.

PR #981, opened 2021-10-11
Update run_jmusicbot script

Prof-Bloodstone:
Changes include:
- Allow overwriting DOWNLOAD and LOOP variables with env vars
- Basic argument parsing support - pass `--no-download` to skip
  downloading, or `--download` to enable it, if previously disabled.
  Same with `--no-loop`.
- Allows passing arguments to the jar after `--`

Example:
```sh
DOWNLOAD=false ./run_jmusicbot.sh --download --no-loop -- --jmusicbot-flag
```
Will cause the download to take place (argument overwrites env var),
without looping, while passing `--jmusicbot-flag` to the process.

### This pull request...
  - [ ] Fixes a bug

### Description
See: https://github.com/jagrosh/MusicBot/pull/650#issuecomment-940463294

This change DOES require bash, it won't work with regular dash.

### DIFF
--- a/scripts/run_jmusicbot.sh
+++ b/scripts/run_jmusicbot.sh
@@ -20,7 +20,7 @@ is_truthy() {
   case "${!varname,,}" in
     yes|true|1) return 0;;
     no|false|0) return 1;;
-    *) printf 'Variable %s has invalid truthy value of %s. Aborting.\n' "${varname}" "${!varname}"; exit 1;;
+    *) printf 'Variable %s is invalid. Expected true/false, got "%s". Aborting.\n' "${varname}" "${!varname}" >&2; exit 1;;
   esac
 }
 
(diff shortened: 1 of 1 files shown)


MichailiK:
Although I haven't tested this, the script looks good to me. jagrosh will have merge this it's all good.

Prof-Bloodstone:
> > If you participate in Hacktoberfest, I'd appreciated getting `hacktoberfest-accepted` label, but it's find if you don't want to :)
> 
> This repository doesn't have the `hacktoberfest` tag, thus it wouldn't count.

It would: https://hacktoberfest.digitalocean.com/resources/participation
> A repository/project is considered to be participating in Hacktoberfest if the 'hacktoberfest' topic is present and is accepting public contributions via pull requests. **An individual pull request can also be opted-in directly by adding the 'hacktoberfest-accepted' label.**

PR #993, opened 2021-10-15
Add PlayskipCmd

Gobidev:
### This pull request...
  - [ ] Fixes a bug

### Description
This pull request adds a playskip command that can be used by DJs. It adds a track to the front of the queue (by calling the playnext command) and then immediately skips to it (by calling the forceskip command) if music is currently playing. This was a feature of the widely used Rythm Bot and I personally use it quite often.
I also added the .idea folder that is being created by the IntelliJ IDEA to the .gitignore.

### Purpose
This feature was the only majour command I missed when migrating from Rythm so I decided to add it.

### Relevant Issue(s)
As much as I am aware, there are no issues associated with this PR.


MichailiK:
Opinion: Wouldn't it be better to call this command `playnow`, since `playnext` exists?

PR #1006, opened 2021-10-18
Convert queueing from reactions to Discord buttons.

SantanaFtRobThomas:
### This pull request...
  - [ ] Fixes a bug

### Description
This removes the old reaction buttons for queueing and replaces them with shiny new Discord buttons.
The new playlist UI:
image
Queue buttons:
image

The `QueueCmd` object now tracks its last message and will remove its buttons from all but the last queue message. If a user tries to navigate to a now-empty queue, this is also dealt with.

As with before, the user is given 30 seconds to accept or decline the addition of a playlist.

### Purpose
Discord buttons are so much more flexible than reactions, and it looks nicer.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/PlayCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/PlayCmd.java
@@ -113,7 +113,7 @@ public class PlayCmd extends MusicCommand {
             AudioHandler handler = (AudioHandler)event.getGuild().getAudioManager().getSendingHandler();
             int pos = handler.addTrack(new QueuedTrack(track, event.getAuthor()))+1;
             String succ = event.getClient().getSuccess();
-            String addMsg = FormatUtil.filter(event.getClient().getSuccess()+"<:jankdacity:837717101866516501> Added **"+track.getInfo().title
+            String addMsg = FormatUtil.filter(event.getClient().getSuccess()+"Added **"+track.getInfo().title
                     +"** (`"+FormatUtil.formatTime(track.getDuration())+"`) "+(pos==0?"to begin playing":" to the queue at position "+pos));
             if(playlist==null || !event.getSelfMember().hasPermission(event.getTextChannel(), Permission.MESSAGE_ADD_REACTION))
                 m.editMessage(addMsg).override(true).queue();


src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
```diff
@@ -177,7 +178,7 @@ else if(config.getGame().getName().equalsIgnoreCase("none"))
                     .setActivity(nogame ? null : Activity.playing("loading..."))
                     .setStatus(config.getStatus()==OnlineStatus.INVISIBLE || config.getStatus()==OnlineStatus.OFFLINE 
                             ? OnlineStatus.INVISIBLE : OnlineStatus.DO_NOT_DISTURB)
-                    .addEventListeners(cb.build(), waiter, new Listener(bot))
+                    .addEventListeners(cb.build(), waiter, new Listener(bot), new QueueButtonListener(settings, cb.build()))
```
jagrosh, line 181: This should not add another listener, it should re-use the EventWaiter

src/main/java/com/jagrosh/jmusicbot/commands/music/QueueCmd.java
```diff
@@ -27,51 +28,54 @@
...
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.Component;
 
 /**
  *
  * @author John Grosh <john.a.grosh@gmail.com>
  */
-public class QueueCmd extends MusicCommand 
-{
+public class QueueCmd extends MusicCommand {
```
jagrosh, line 47: This is the first instance of many where you have converted next-line braces to same-line braces. JMusicBot uses next-line.
```java
void someMethod()
{
    // content
}
```

src/main/java/com/jagrosh/jmusicbot/commands/music/QueueCmd.java
```diff
@@ -86,6 +90,32 @@ public void doCommand(CommandEvent event)
             });
             return;
         }
+
+        if(pagenum > ah.getQueue().getNumberOfPages()){
+            event.reply("That page does not exist. The maximum number of pages on the queue is " + String.valueOf(ah.getQueue().getNumberOfPages()) + ". Showing last page.");
```
jagrosh, line 95: We shouldn't be sending an extraneous message when we're not changing the behavior _and_ the changed behavior is already visible (in the different page number).

src/main/java/com/jagrosh/jmusicbot/queue/FairQueue.java
```diff
@@ -20,6 +20,8 @@
 import java.util.List;
 import java.util.Set;
 
+import net.dv8tion.jda.api.entities.Message;
```
jagrosh, line 23: FairQueue is a very generic container and should definitely not contain any references to JDA, JMusicBot, or any music-related systems at all. Do not modify this class unless it is to add generic functionality.

jagrosh (changes requested):
So... while I like the _idea_ of this change, this implementation is not at all how I would have done it. There's a ton of formatting changes, and a lot of changes to core systems that are not meant to be changed in the way you've changed them. Frankly, the best way to go about adding this functionality is to create a class that extends the normal Paginator, but listens for button events instead of reaction adds.

I'm not going to close this right now, but I'm also not sure if there is anything in this PR that would be salvageable for a PR doing this in a way that more matches how JMusicBot is set up.

Issue #353, opened 2020-02-10
[Feature Request] Add a setting for automatic "Now playing" information when a song starts playing


HyperGamer14:
**Is your feature request related to a problem? Please describe.**
When a song starts playing the bot doesn't automatically show which one. That can be a little bit annoying when doing something else while the bot is playing in the background.

**Describe the solution you'd like**
When the bot starts a new song, it automatically posts the song title similar to the "Now playing" command

**Describe the basic flow/steps of using this feature**
How this feature would be used:
1. Turn on the feature via a command like "!setautonowplaying"
2. Bot starts song
3. Bot posts "Now playing" information (Title, song length)

jagrosh:
Although not quite the exact feature you're describing, the `nowplaying` command already continues to update - including when new songs start - assuming you have `npimages = false` in your config.

johnb-xp:
If this was to implemented, being able to choose a "now playing" channel would be nice to reduce spam lol

pierluigi:
I have issued `!nowplaying` however it only responds once – when the next song starts I don't get the message in the chat (only in the bot's own "Listening to ... " status message and in the text chat channel description.

NitroXGamer1235:
my bot keeps saying on startup :
internal/modules/cjs/loader.js:983
  throw err;
  ^

Error: Cannot find module './config.json'
Require stack:
- C:\Users\Kids\Documents\Music Bot\index.js
[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:980:15)[39m
[90m    at Function.Module._load (internal/modules/cjs/loader.js:862:27)[39m
[90m    at Module.require (internal/modules/cjs/loader.js:1042:19)[39m
[90m    at require (internal/modules/cjs/helpers.js:77:18)[39m
    at Object.<anonymous> (C:\Users\Kids\Documents\Music Bot\index.js:2:27)
[90m    at Module._compile (internal/modules/cjs/loader.js:1156:30)[39m
[90m    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)[39m
[90m    at Module.load (internal/modules/cjs/loader.js:1000:32)[39m
[90m    at Function.Module._load (internal/modules/cjs/loader.js:899:14)[39m
[90m    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)[39m {
  code: [32m'MODULE_NOT_FOUND'[39m,
[...]

PR #1010, opened 2021-10-23
Automatically show "now playing" when a new track starts

briankendall:
### This pull request...
  - [ ] Fixes a bug

### Description
This PR adds a bot setting `autonowplaying` that will make the bot automatically send a "now playing" text message when a new track starts, for any guild that has configured a text channel using the `settc` command. The setting defaults to off.

If `npimages` is true then the now playing message will not include a progress bar since it will always be at 00:00 and will never update.

### Purpose
For communities that want to see what the currently playing track is in their text channel's messages, similar to other music bots.

This provides another mechanism for users to easily see what the currently playing track is and, in particular, when the track changes. Aside from users preferring this method, this could also be useful for use cases where it's not possible to set the bot's status or not feasible to set the channel's topic (for example due to not having permission from the server or due to Discord's rate limiting).

### Relevant Issue(s)
#353


MichailiK:
While skimming, I noticed that this is a bot/config setting, not a server setting. I believe that this would much rather fit as a per-server setting.

PR #1047, opened 2021-11-19
ft: enqueue every file attachement, not just the first

Mattwmaster58:
### This pull request...
  - [ ] Fixes a bug

### Description
When uploading files, usually only the first file is considered. This changes that. I've never used Java before so I'm extra open to criticism on this PR.

### Purpose
Multiqueueing uploads is a convenience thing

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/PlayCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/PlayCmd.java
@@ -31,6 +31,9 @@ import com.jagrosh.jmusicbot.commands.MusicCommand;
 import com.jagrosh.jmusicbot.playlist.PlaylistLoader.Playlist;
 import com.jagrosh.jmusicbot.utils.FormatUtil;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import net.dv8tion.jda.api.Permission;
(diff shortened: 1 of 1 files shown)


jagrosh:
I have some initial concerns about this PR:

1. One of the next things that we need to re-write is the result handler. Right now, there is a lot of duplicated code across many commands, and it would be useful to consolidate everything into a common result handler that can take additional information about the request and handle all additions to the queue, regardless of which command they came from.
2. There are a few additional implications of this idea of queuing multiple tracks. For example, if you can queue multiple tracks in one command via uploads, shouldn't you also be able to queue multiple tracks via multiple URLs as well? I don't want to add partial-functionality, so that feature would need to exist in this PR as well.

Mattwmaster58:
WRT 1: I'm not quite sure what you mean here, but in general can see what you mean WRT duplicated code in general. I'm not experienced in Java at all so I'm not sure how I'd be able to tackle this well.

2: I agree, and this is what I started to try and implement in eg 87ac8e3, but the way the queuing system works it's not trivial. I was a bit hasty in bringing this from draft to ready for review but upon somewhat understanding this I converted back to a draft. Anything I come up architecturally will probably be sub-par as I'm really not experienced in Java, though

Mattwmaster58:
Anyone feel free to pick up where I left off, but I no longer plan to work on this PR

PR #1055, opened 2021-11-26
Minor enhancement

yo-louk:
### This pull request...
  - [ ] Fixes a bug

### Description
In this PR I have made very minor changes, improving the overall grammar of the bot's responses.
Here are a few examples, version 0.3.6 being on top and this PR's below.

`1 entries` → `1 entry`
image
image

`1 items` → `1 item`
image

`1 entries` → `1 entry`
image
image

### Purpose
Saying `1 entry` instead of `1 entries` just makes more sense. (grammar duh)
This seemed like a small thing I thought I would be able to fix.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/dj/ForceRemoveCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/dj/ForceRemoveCmd.java
@@ -113,7 +113,7 @@ public class ForceRemoveCmd extends DJCommand
         }
         else
         {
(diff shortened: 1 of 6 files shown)


yo-louk:
Let me know if I missed a spot.

jagrosh:
While this does resolve some grammar issues, I think that everything in this PR will end up getting overwritten when we add localization, which will utilize resource bundles and do plurals like this: https://docs.oracle.com/javase/tutorial/i18n/format/choiceFormat.html

Issue #982, opened 2021-10-12
[Feature Request] Skip non-music sections of music video tracks


Brando47:
### Is your feature request related to a problem? Please describe.

Music videos from youtube are often selected as the top result, containing long non-music sections.

### What is your ideal solution to the problem?

"SponsorBlock is an open-source crowdsourced browser extension and open API for skipping sponsor segments in YouTube videos", see [https://sponsor.ajay.app/](https://sponsor.ajay.app/). In particular, one of the categories is non-music sections of music videos, see [https://wiki.sponsor.ajay.app/w/Music:_Non-Music_Section](https://wiki.sponsor.ajay.app/w/Music:_Non-Music_Section).

In other words, it would use crowdsourced data as a means of choosing segments of the track to skip such that only music containing segments are played.

It seems like a really nice fit for this project.

### How would this feature be used?

I would suggest an opt-in setting to enable the feature, perhaps that could be activated by DJ users as a bot command. Perhaps additionally there could also be a command that can be used during playback to skip a specific segment as it's playing.

### Checklist

- [X] I have checked [the documentation](https://jmusicbot.com/) to make sure this feature doesn't already exist
- [X] I have searched for [similar feature requests](https://github.com/jagrosh/MusicBot/labels/enhancement)
- [X] I have confirmed that my requests is **not** on the [list of things that will not be added](https://jmusicbot.com/things-that-wont-be-added)
- [X] I am running the latest version of the bot: ![Release](https://img.shields.io/github/release/jagrosh/MusicBot.svg)

PR #1067, opened 2021-12-06
Added skipsegment command that interacts with SponsorBlock API to skip non music segments on YouTube music videos

Tommy-Davies:
### This pull request...
  - [ ] Fixes a bug

### Description
This implements feature request #982 and adds a new command that allows a user to optionally skip the non-music segment that is currently playing. This feature is implemented with the skipSegCmd class which retrieves the URL of the currently playing song, extracts the video ID, sends a request to the SponsorBlock API, and parses the JSON file. The JSON parser returns either an error code (-1) or the millisecond value of the end of the current segment if one is found. If it succeeded, it will tell the user where the track is seeking to. If it failed it will tell the user why. It uses the lava player setPosition() method to set the track position to the correct time upon success.

### Purpose
When listening to a music video on YouTube, there are often non-music sections or intros that are impossible to skip, and can be very frustrating to sit through without visual context. This often makes these versions of songs unplayable. This feature is invoked with the "skipsegment" command. The user will use this command while a song is playing to skip to the end of the current segment in a song. The command will return the timestamp for the segment that it is jumping ahead to or display “No segments found”. If the song is in the last segment of the song and the user tries to skipSegment, the message “Cannot skip here because no segment is currently playing!”.  This happens since there are no segments left in the song. This feature only works with YouTube videos, due to the limitations of the SponsorBlock API, so if the user tries to skipSegment while playing a SoundCloud, Vimeo, or Twitch stream, they will get a “Could not get video ID!” message. 

Command succeeding by skipping a segment that is currently playing and moving the track ahead to the correct location:
Screen Shot 2021-12-05 at 8 36 49 PM

Command succeeding at retrieving database entry but failing because no non-music segment is currently playing:
Screen Shot 2021-12-05 at 8 36 59 PM

Command failing because the song is not in the SponsorBlock database:
Screen Shot 2021-12-05 at 8 37 10 PM

### Relevant Issue(s)
This PR closes feature request #982

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -99,7 +99,7 @@ public class JMusicBot
                         new PingCommand(),
                         new SettingsCmd(bot),
                         
-                        new LyricsCmd(bot),
+                        // new LyricsCmd(bot),
                         new NowplayingCmd(bot),
                         new PlayCmd(bot),
                         new PlaylistsCmd(bot),
@@ -109,6 +109,7 @@ public class JMusicBot
                         new SCSearchCmd(bot),
                         new ShuffleCmd(bot),
                         new SkipCmd(bot),
+                        new skipSegCmd(bot),
 
                         new ForceRemoveCmd(bot),
                         new ForceskipCmd(bot),
--- a/src/main/java/com/jagrosh/jmusicbot/commands/music/skipSegCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/music/skipSegCmd.java
@@ -1,0 +1,177 @@
+package com.jagrosh.jmusicbot.commands.music;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import com.jagrosh.jdautilities.command.CommandEvent;
+import com.jagrosh.jmusicbot.Bot;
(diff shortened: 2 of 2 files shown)


src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
```diff
@@ -99,7 +99,7 @@ public static void main(String[] args)
                         new PingCommand(),
                         new SettingsCmd(bot),
                         
-                        new LyricsCmd(bot),
+                        // new LyricsCmd(bot),
```
Tommy-Davies, line 102: This line was causing problems when I tried to run the bot from a jar executable.

src/main/java/com/jagrosh/jmusicbot/commands/music/skipSegCmd.java
```diff
@@ -0,0 +1,177 @@
...
+     * @throws IOException
+     */
+    public long connectToAPI(String videoID, AudioHandler handler, CommandEvent event)
+            throws UnsupportedEncodingException, IOException {
+        HttpURLConnection con = null;
+        long curTime = getCurrentTimeStamp(handler) / 1000; // convert to seconds
+        boolean musicSegment = false;
+
+        try {
+            String nonMusicURLString = "https://sponsor.ajay.app/api/searchSegments?videoID=";
```
Joe-Dowd, line 109: Should use the skipSegments endpoint rather than searchSegments ([docs](https://wiki.sponsor.ajay.app/w/API_Docs#GET_.2Fapi.2FsearchSegments)). Using the [hash prefix version of skipSegments](https://wiki.sponsor.ajay.app/w/API_Docs#GET_.2Fapi.2FskipSegments.2F:sha256HashPrefix) would be better for user privacy.

Also, `category` should be supplied as a query parameter with the value of `music_offtopic` (or `categories` if multiply types are to be skipped) to ensure you're getting non-music segments back, as the default is to only retrieve `sponsor` segments. ([docs](https://wiki.sponsor.ajay.app/w/Types))

Issue #1122, opened 2022-02-24
[Bug Report] bot can't play shorts video


Megumiso:
### Bug Description

bot can't play shorts video.

This sounds like a really simple issue since YouTube Shorts were released few years ago, but I couldn't find a same issue by searching "short" or "shorts". so I'm reporting this.
Sorry if there's already the same issue that I missed.

### Steps to Reproduce

1. !play https://www.youtube.com/shorts/V6Qipms13VE
2. bot says "Error loading: Unknown file format."

### Expected Result

bot starts playing shorts video just like usual YouTube video.

### Debug Output

```
System Properties:
  java.version = 11.0.13
  java.vm.name = OpenJDK 64-Bit Server VM
  java.vm.specification.version = 11
  java.runtime.name = OpenJDK Runtime Environment
  java.runtime.version = 11.0.13+8-Ubuntu-0ubuntu1.20.04
  java.specification.version = 11
  os.arch = amd64
  os.name = Linux

JMusicBot Information:
  Version = 0.3.6
  Owner = 241578268749201408
  Prefix = !
  AltPrefix = null
  MaxSeconds = 0
  NPImages = false
  SongInStatus = false
  StayInChannel = true
  UseEval = false
  UpdateAlerts = true

Dependency Information:
  JDA Version = 4.3.0_324
  JDA-Utilities Version = 3.0.5
  Lavaplayer Version = 1.3.78

Runtime Information:
  Total Memory = 17
  Used Memory = 13

Discord Information:
  ID = 418666605262798848
  Guilds = 12
  Users = 5
```

### Additional Info

You can workaround this issue by changing the URL like following:

 https://www.youtube.com/shorts/V6Qipms13VE
-> https://www.youtube.com/watch?v=V6Qipms13VE

### Checklist

- [X] I have looked for information about this [within the documentation](https://jmusicbot.com/)
- [X] I have searched for similar issues [on the issues page](https://github.com/jagrosh/MusicBot/issues?q=is%3Aissue)
- [X] I am running the latest version of the bot: ![Release](https://img.shields.io/github/release/jagrosh/MusicBot.svg)

Issue #885, opened 2021-09-20
[Bug Report] Error loading track from YouTube (403)


MichailiK:
### Bug Description

Sometimes, a track fails to play & will get skipped without any prompt. The following exception will be thrown, indicating JMusicBot received a 403 response from YouTube:

<details>

<summary>Log output</summary>

```
Error in playback of Wg92RrNhB8s
com.sedmelluq.discord.lavaplayer.tools.FriendlyException: Something broke when playing the track.
        at com.sedmelluq.discord.lavaplayer.tools.ExceptionTools.wrapUnfriendlyExceptions(ExceptionTools.java:44)
        at com.sedmelluq.discord.lavaplayer.track.playback.LocalAudioTrackExecutor.execute(LocalAudioTrackExecutor.java:116)
        at com.sedmelluq.discord.lavaplayer.player.DefaultAudioPlayerManager.lambda$executeTrack$1(DefaultAudioPlayerManager.java:348)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
        Suppressed: com.sedmelluq.discord.lavaplayer.tools.exception.EnvironmentInformation:
  lavaplayer.version: 1.3.78
  os.arch: amd64
  os.name: Linux
  os.version: 4.18.0-305.7.1.el8_4.x86_64
  java.vendor: Red Hat, Inc.
  java.version: 11.0.11
  java.runtime.version: 11.0.11+9-LTS
  java.vm.version: 11.0.11+9-LTS
Caused by: java.lang.RuntimeException: Not success status code: 403
        at com.sedmelluq.discord.lavaplayer.tools.io.PersistentHttpStream.validateStatusCode(PersistentHttpStream.java:87)
        at com.sedmelluq.discord.lavaplayer.tools.io.PersistentHttpStream.attemptConnect(PersistentHttpStream.java:116)
        at com.sedmelluq.discord.lavaplayer.tools.io.PersistentHttpStream.connect(PersistentHttpStream.java:105)
        at com.sedmelluq.discord.lavaplayer.tools.io.PersistentHttpStream.internalRead(PersistentHttpStream.java:150)
[...]

PR #1126, opened 2022-02-28
Fix youtube

Sartharon:
### This pull request...
  - Fixes a bug

### Description
Fixes [#885](https://github.com/jagrosh/MusicBot/issues/885) by using a [forked](https://github.com/Walkyst/lavaplayer-fork) version of lavaplayer since the original seems dead atm.
This PR also adds optional support for the needed tokens if a youtube video is truely an age restricted video. Documentation on how to retireve the correct tokens can be provided in a PR as well. (Yes I'm well aware that this project aims to minimize the need for tokens but it seems to me that this is a too common issue to be left alone imo.)

An upcoming version will also fix [#1122](https://github.com/jagrosh/MusicBot/issues/1122)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.jagrosh</groupId>
     <artifactId>JMusicBot</artifactId>
-    <version>Snapshot</version>
+    <version>0.3.6.1</version>
     <packaging>jar</packaging>
     
     <repositories>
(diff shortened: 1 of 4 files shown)


MichailiK:
We are using our [own fork of lavaplayer](https://github.com/jagrosh/lavaplayer/tree/jmusicbot). Please suggest or PR any patches there instead.

FlyGuyGo:
Thank you for making this @Sartharon [Sartharon](https://github.com/Sartharon)! I often encounter error 403, so I am glad to be able to use this as a stop-gap solution.

adamhradil:
Thank you @Sartharon!

PR #1139, opened 2022-03-31
Allow using the -stop command without a DJ role set

t3rminus:
### This pull request...
  - [ ] Fixes a bug
  - [X] Improves an existing feature

### Description
If no DJ role is set, any user can now stop the bot and remove it from the channel. 

### Purpose
This is very useful for well-behaved servers, where admins haven't configured or don't want to configure DJ roles. This prevents someone from accidentally leaving a large music queue running, and leave the voice channel, and regular users have no way of stopping the bot.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/DJCommand.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/DJCommand.java
@@ -43,6 +43,6 @@ public abstract class DJCommand extends MusicCommand
             return true;
         Settings settings = event.getClient().getSettingsFor(event.getGuild());
         Role dj = settings.getRole(event.getGuild());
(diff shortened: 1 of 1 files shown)


jagrosh:
If you intend for all users to use DJ commands, then set the DJ role to @ everyone

t3rminus:
> If you intend for all users to use DJ commands, then set the DJ role to @ everyone

Why not make it a default? It seems odd to gate half the bot's commands behind a non-obvious configuration setting.

jagrosh:
It's an elevated permission, it shouldn't be enabled by default, and it's very easy to make the change when you're setting up the bot.

jagrosh:
For example, someone not knowing how to disable it is far more destructive than someone not knowing how to enable it.

PR #1140, opened 2022-04-01
Allow configurable default skip ratio

t3rminus:
### This pull request...
  - [ ] Fixes a bug

### Description
Allows the server owner to configure the skip ratio in config.txt, including disabling vote-related messages if skip ratio is 0

### Purpose
Useful for owners being able to set a default ratio. Setting to "0" and having no vote-related messages is also helpful for well-behaved, or small servers, where music playing is mostly co-operative, and people may want to skip troll songs or livestreams that have gone on too long.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
+++ b/src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
@@ -15,6 +15,7 @@
  */
 package com.jagrosh.jmusicbot.settings;
 
+import com.jagrosh.jmusicbot.BotConfig;
 import com.jagrosh.jdautilities.command.GuildSettingsManager;
 import com.jagrosh.jmusicbot.utils.OtherUtil;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
```diff
@@ -81,7 +84,7 @@ public Settings getSettings(long guildId)
     
     private Settings createDefaultSettings()
     {
-        return new Settings(this, 0, 0, 0, 100, null, RepeatMode.OFF, null, SKIP_RATIO);
+        return new Settings(this, 0, 0, 0, 100, null, RepeatMode.OFF, null, this.config.getSkipRatio());
```
MichailiK, line 87: This should be updated to use a negative number.

src/main/java/com/jagrosh/jmusicbot/settings/SettingsManager.java
```diff
@@ -31,12 +32,14 @@
  */
 public class SettingsManager implements GuildSettingsManager
 {
-    private final static double SKIP_RATIO = .55;
     private final HashMap<Long,Settings> settings;
+    private final BotConfig config;
 
-    public SettingsManager()
+    public SettingsManager(BotConfig config)
```
MichailiK, line 38: Any references to the `BotConfig` in this class should become unnecessary, once my suggestions are applied. Please remove them.

MichailiK (changes requested):
As this currently stands, this can lead to behavior which users might not expect, when changing the skip ratio in the config. I would suggest to instead use negative numbers (like -1) to indicate that a guild has not specified a skip ratio. The `skip` command could then determine whether to use the skip ratio of the guild or the bot.

MichailiK:
Lastly, [`src/main/resources/reference.conf`](https://github.com/jagrosh/MusicBot/blob/master/src/main/resources/reference.conf) needs to be updated, adding the new option to the bot config.

t3rminus:
I've made the changes you suggested.

MichailiK (approved):
Whoops, I didn't notice I haven't reviewed & approved this. My apologies for that! Jagrosh will have another look at this, I'm hoping we can get this merged & into the next release very soon.

PR #1165, opened 2022-04-30
Another type of queue

SH1n3g4ter:
### This pull request...
  - [ ] Fixes a bug

### Description
This adds another type of queue, instead of the already existing `FairQueue`.
This `UnfairQueue` strictly maintains insertion order.
Songs are just consecutively added to the list independent of who added them.

I also wrote an interface `Queue` to generalize the queue types.

You can change which type to use by setting the **config.txt** option `queuealgorithm` to either `"fair"` or `"unfair"`.
The default queue type remains `FairQueue` and continues to function as usual.

### Purpose
When multiple users add different playlists the songs get mixed up due to the `FairQueue`, which might be wanted or not.
I personally prefer it not to be like that, thats why I added another queue type. 

### Relevant Issue(s)
None

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
+++ b/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
@@ -40,7 +40,7 @@ public class BotConfig
     
     private Path path = null;
     private String token, prefix, altprefix, helpWord, playlistsFolder,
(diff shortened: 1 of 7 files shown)


jagrosh (changes requested):

MichailiK:
Closing this in favor of #1194

PR #1230, opened 2022-10-09
Add message content intent step

MichailiK:
Adds a step to enable Message Content Intent with this image:

Privileged intents


jagrosh:
If we're going to add this, let's include the section about making sure it's a private bot as well. We should also check the Guild Members intent in case we ever have a feature that needs that at some point in the future.

jagrosh:
image

Issue #1303, opened 2023-02-27
[Bug Report] No handling for no internet connection


LachlanPond:
### Bug Description

When running the bot without an internet connection, the JDA connection attempt will fail without catching the exception and doing a proper system exit. When a server (physical PC) running the bot restarts, if there is no internet connection when the system starts, the bot will fail without crashing and therefore cannot be restarted automatically.

`[06:02:09] [ERROR] [Requester]: DNS resolution failed: discord.com: Name or service not known
Exception in thread "main" net.dv8tion.jda.api.exceptions.ErrorResponseException: -1: java.net.UnknownHostException
        at net.dv8tion.jda.internal.requests.RestActionImpl.complete(RestActionImpl.java:227)
        at net.dv8tion.jda.api.requests.RestAction.complete(RestAction.java:633)
        at net.dv8tion.jda.api.utils.SessionControllerAdapter.getGateway(SessionControllerAdapter.java:86)
        at net.dv8tion.jda.internal.JDAImpl.getGateway(JDAImpl.java:318)
        at net.dv8tion.jda.internal.JDAImpl.login(JDAImpl.java:276)
        at net.dv8tion.jda.internal.JDAImpl.login(JDAImpl.java:268)
        at net.dv8tion.jda.api.JDABuilder.build(JDABuilder.java:1874)
        at com.jagrosh.jmusicbot.JMusicBot.startBot(JMusicBot.java:187)
        at com.jagrosh.jmusicbot.JMusicBot.main(JMusicBot.java:65)
Caused by: java.net.UnknownHostException: discord.com: Name or service not known

...

Caused by: net.dv8tion.jda.api.exceptions.ContextException
        at net.dv8tion.jda.api.exceptions.ContextException.here(ContextException.java:54)
        at net.dv8tion.jda.api.requests.Request.<init>(Request.java:71)
        at net.dv8tion.jda.api.requests.RestFuture.<init>(RestFuture.java:36)
        at net.dv8tion.jda.internal.requests.RestActionImpl.submit(RestActionImpl.java:209)
        at net.dv8tion.jda.internal.requests.RestActionImpl.complete(RestActionImpl.java:219)
        at net.dv8tion.jda.api.requests.RestAction.complete(RestAction.java:633)
        at net.dv8tion.jda.api.utils.SessionControllerAdapter.getGateway(SessionControllerAdapter.java:86)
        at net.dv8tion.jda.internal.JDAImpl.getGateway(JDAImpl.java:318)
[...]

PR #1304, opened 2023-02-27
Catch JDA exception when there is no internet connection

LachlanPond:
### This pull request...
  - [x] Fixes a bug

### Description
Added catch clause to JMusicBot.java

### Purpose
When running the bot without an internet connection, the JDA connection attempt will fail without catching the exception and doing a proper system exit. This means that if the bot cannot establish a connection, the program will wait instead of exiting. Automated restart scripts will not work as a result.

### Relevant Issue(s)
This PR closes issue #1303


Traydr:
It's missing `import net.dv8tion.jda.api.exceptions.ErrorResponseException` for the `ErrorResponseException`.

LachlanPond:
> It's missing `import net.dv8tion.jda.api.exceptions.ErrorResponseException` for the `ErrorResponseException`. Not the author, but I noticed it yesterday.

Yeah that's my mistake, I've added it in and tested on my machine. I have a systemd service that will now automatically restart on failure.

PR #1332, opened 2023-03-21
Made update check optional

t3rminus:
### This pull request...
  - [ ] Fixes a bug

### Description
Adds an option to the config file to make update checking optional

### Purpose
Allows additional features to be configured, giving the bot owner more control over how the bot operates.


jagrosh:
Why is this needed?

t3rminus:
Why not? Always good to add configurability, in my opinion.

jagrosh:
I think it's more important to only add important and useful features to reduce clutter and confusion, and as far as I can tell, this addition doesn't add any useful functionality. One request to GitHub hardly hurts anything (especially considering the massive number of requests that bots are making to Discord), and the request itself is entirely surrounded by try-catch so even in the case that GitHub were unreachable, the bot wouldn't crash.

PR #1363, opened 2023-04-27
Add custom command to launch frequently used URL

Wobak:
### This pull request...
  - [ x ] Introduces a new feature

### Description
This commands implements the <prefix>custom command which allows to use the shortcuts config to run commands such as 
!custom mymusic to avoid using the full URL every time.

### Purpose
When using the same music multiple times, it might be useful to add it in the config for custom calls

### Relevant Issue(s)
None noted


MichailiK:
You seem to have changed every file's mode/permission in the project, thus creating a diff out of every file. You also unintentionally committed [binaries](https://github.com/jagrosh/MusicBot/tree/e2303ecfacfd541b781d8755fbdceeb12bd8648c/src/main/resources/natives) and other unwanted files into your branch. Please undo those, then force-push your changes into the branch.

Wobak:
Hello,

OK let me try to review that.

However the binaries are now in the main project as part of last commit : https://github.com/jagrosh/MusicBot/tree/master/src/main/resources/natives

MichailiK:
Oh. I apologize, I never noticed they were added. I got confused, since every file is being shown as modified in this PR. It'd be great if all the files can be reset back to how they are in our master branch.

Wobak:
We added 2 commits with the reset of permissions and the new code logic. Let me know if this is good or not :)

Wobak:
I'm changing a few steps for the integration to be more consistent with the rest of the project.

Wobak:
@jagrosh, any chance this can make it to the repo? I don't mind if you don't want to, just refuse the PR so that I know :)

Thanks again for the hard work !

Wobak:
This is now broken following the update on Lavacommand & so on.

Issue #1322, opened 2023-03-12
[Feature Request] Blacklist a user


snek7986:
### Is your feature request related to a problem? Please describe.

I have a user in my discord who keeps using !stop and I can't blacklist them. I have to have the dj role as the normal member role as I would like users to be able to control the volume.

### What is your ideal solution to the problem?

To add a !blacklist command that you can use with the users user id to block them from using commands.

### How would this feature be used?

1. get the id of the user
2. type (prefix)blacklist (id)

### Checklist

- [X] I have checked [the documentation](https://jmusicbot.com/) to make sure this feature doesn't already exist
- [X] I have searched for [similar feature requests](https://github.com/jagrosh/MusicBot/labels/enhancement)
- [X] I have confirmed that my requests is **not** on the [list of things that will not be added](https://jmusicbot.com/things-that-wont-be-added)
- [X] I am running the latest version of the bot: ![Release](https://img.shields.io/github/release/jagrosh/MusicBot.svg)

hereticerik:
I would really like this feature as well, this would be a very useful feature.

PR #1367, opened 2023-05-02
Blacklist user from using music and DJ commands

glizzygobblersupreme:
### This pull request...
  - [ ] Fixes a bug

### Description
Adds command to blacklist a user from using music and DJ commands
`!blacklist <user>` - adds or removes user from blacklist
`!blacklist none` - clear all users from blacklist

This allows the admin to continue having all members as DJs while allowing them to prevent certain users from abusing music controls

### Relevant Issue(s)
This PR closes issue #1322

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/admin/BlacklistUserCmd.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/admin/BlacklistUserCmd.java
@@ -17,14 +17,19 @@ package com.jagrosh.jmusicbot.commands.admin;
 
 import com.jagrosh.jdautilities.command.CommandEvent;
 import com.jagrosh.jdautilities.commons.utils.FinderUtil;
+import com.jagrosh.jdautilities.menu.OrderedMenu;
 import com.jagrosh.jmusicbot.Bot;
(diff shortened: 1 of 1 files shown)


src/main/java/com/jagrosh/jmusicbot/commands/admin/BlacklistUserCmd.java
```diff
@@ -0,0 +1,104 @@
...
+        User target;
+        List<Member> found = FinderUtil.findMembers(event.getArgs(), event.getGuild());
+        if(found.isEmpty())
+        {
+            event.replyError("Unable to find the user!");
+            return;
+        }
+        else if(found.size()>1)
+        {
+            StringBuilder builder = new StringBuilder();
```
mnasiatka, line 76: I went with `StringBuilder` instead of `OrderedMenu.Builder` to display duplicates because it seems like any user would be able to interact with the emoji options and blacklist the incorrect user, which is not the intended use case. If `OrderedMenu.Builder` checks that the author is the one choosing the option I think it'd be fine to switch back to it though

Matdbx10:
its a good idea

PR #1397, opened 2023-06-28
Support new Discord username

MagicTeaMC:
### This pull request...
  - [x] Fixes a bug

### Description
A easy way to support Discord new username(maybe

### Purpose
圖片
圖片

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/audio/AudioHandler.java
+++ b/src/main/java/com/jagrosh/jmusicbot/audio/AudioHandler.java
@@ -214,15 +214,17 @@ public class AudioHandler extends AudioEventAdapter implements AudioSendHandler
             {
                 User u = guild.getJDA().getUserById(rm.user.id);
                 if(u==null)
-                    if(rm.user.discrim.equals("0000"))
+                    if(rm.user.discrim.equals("0000")) {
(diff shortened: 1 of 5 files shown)


jagrosh:
I don't like how this is implemented; I think we should add a formatting function somewhere that will handle the logic, and then use the formatted output anywhere we need the username

MagicTeaMC:
discord will force all user to use new username, so we don't need to check username anymore :)

jagrosh:
Either way I think it would still be good to make a specific formatting function for usernames. Discrims will still exist for a bit longer (this would be a future-proof solution) and then it becomes easier to make other formatting changes in the future (for example, showing display name instead of username, or showing a combination of them).

MichailiK (changes requested):
The new `formatUsername` utility is missing usage in 2 places:

https://github.com/jagrosh/MusicBot/blob/8457f944cdaada4f37f0c305624d0b292a2b6043/src/main/java/com/jagrosh/jmusicbot/commands/dj/ForceskipCmd.java#L45

https://github.com/jagrosh/MusicBot/blob/8457f944cdaada4f37f0c305624d0b292a2b6043/src/main/java/com/jagrosh/jmusicbot/commands/music/SkipCmd.java#L69

MichailiK (approved):

PR #1425, opened 2023-08-18
Allow for changing systemsettings file name

KamilKurde:
### This pull request...
  - [ ] Fixes a bug

### Description
This pull request adds a possibility to change `serversettings` filename

### Purpose
Making it easier to host multiple bot instances, by eliminating the need for them to have their own working directories with `serversettings.json` that's not going to be modified by other instances


jagrosh:
This was discussed briefly in Discord and this is a feature that we want added, but it should be set from the config file instead of a system property. Using the config won't prevent you from using a system property (as HOCON can read system properties and environment variables)

PR #1486, opened 2024-02-28
Make JMusicBot error & shut down if ran in an unsupported manner

MichailiK:
This PR adds a `getUnsupportedBotReason` utility, which determines if the bot JMusicBot is running under shows any remarks of being run in a manner that JMusicBot doesn't support. JDA will get shut down & JMusicBot will exit if there is a reason.

Right now, it only checks:

- If the bot is verified
- If the bot is public

It doesn't look like there's a way to check privileged intents with JDA 4? Thus its absent for now.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -187,6 +187,14 @@ public class JMusicBot
                     .setBulkDeleteSplittingEnabled(true)
                     .build();
             bot.setJDA(jda);
+
+            String unsupportedReason = OtherUtil.getUnsupportedBotReason(jda);
+            if (unsupportedReason != null)
+            {
+                jda.shutdown();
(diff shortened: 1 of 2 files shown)


MichailiK:
Whoops, the GUI prompt immediately disappears when JDA shuts down. Not exactly sure how to tackle that right now, so any suggestions would be welcome!

tbodt:
I'm confused about why we have this, my bot is set to public and is working fine, at least as far as I can tell.

sks316:
I'm confused as well. Until the latest update, my bot was set to public with absolutely no issues, then when I update, it stops working. I don't see the rationale behind this.
Fun fact, if the bot is set to auto start as a systemd service and meets one of these "unsupported" conditions, this new update causes it to hit the Discord API over 1,000 times in a short period, which makes Discord invalidate the bot token due to "potential abuse". That's the only way I knew anything was wrong, because I usually don't check the systemd logs after updating this.

tbodt:
I just turned off update notifications - this one doesn't seem worth it.

PR #1487, opened 2024-02-28
Don't warn about failing to load server settings when the file isn't found

MichailiK:
Users get confused by the `Failed to load server settings (this is normal if no settings have been set yet)` warning, which almost always gets logged when the serversettings.json file doesn't exist. The file not existing is perfectly fine for JMusicBot, so I propose we don't log this when the file wasn't found.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
+++ b/src/main/java/com/jagrosh/jmusicbot/JMusicBot.java
@@ -165,8 +165,10 @@ public class JMusicBot
                 GUI gui = new GUI(bot);
                 bot.setGUI(gui);
                 gui.init();
-            } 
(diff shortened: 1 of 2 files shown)


jagrosh:
Can we make and info-level statement regardless of if it loads or not:
`server-settings.json loaded from <filename>`
`server-settings.json will be created in <filename where it was not found>`

MichailiK:
`serversettings.json` will now be created if it the file doesn't exist. It'll also log when successfully loading it as well.

> Also we need to add an info statement about the location of the config file that prints after the gui is open

I've done this by logging the exact same `Loaded config from <path>` message again after successfully initializing the GUI. It works, feels a bit hacky though. If there's a better way to implement this please mention it!

Issue #1501, opened 2024-03-09
[Bug Report] Show a relevant error message when trying to use eval with an invalid script enging


ramboozer:
### Bug Description

Eval not working
Version 0.4.0

### Steps to Reproduce

DM or in channel
.eval jda.guilds

### Expected Result

I expect it to list the servers it in currently in.

### Debug Output

`System Properties:
  java.version = 20.0.1
  java.vm.name = Java HotSpot(TM) 64-Bit Server VM
  java.vm.specification.version = 20
  java.runtime.name = Java(TM) SE Runtime Environment
  java.runtime.version = 20.0.1+9-29
  java.specification.version = 20
  os.arch = amd64
  os.name = Windows 10

JMusicBot Information:
  Version = 0.4.0
  Owner = 
  Prefix = .
  AltPrefix = null
  MaxSeconds = 0
  NPImages = false
  SongInStatus = true
  StayInChannel = true
  UseEval = true
  UpdateAlerts = true

Dependency Information:
  JDA Version = 4.4.1_353
  JDA-Utilities Version = 3.0.5
  Lavaplayer Version = 727959e9f621fc457b3a5adafcfffb55fdeaa538-SNAPSHOT

Runtime Information:
  Total Memory = 40
  Used Memory = 23

Discord Information:
  ID = 
  Guilds = 3
  Users = 18`

### Additional Info

[16:40:19] [ERROR] [JDA]: One of the EventListeners had an uncaught exception
java.lang.NullPointerException: Cannot invoke "javax.script.ScriptEngine.put(String, Object)" because "se" is null
	at com.jagrosh.jmusicbot.commands.owner.EvalCmd.execute(EvalCmd.java:46)
	at com.jagrosh.jdautilities.command.Command.run(Command.java:323)
	at com.jagrosh.jdautilities.command.impl.CommandClientImpl.onMessageReceived(CommandClientImpl.java:557)
	at com.jagrosh.jdautilities.command.impl.CommandClientImpl.onEvent(CommandClientImpl.java:445)
	at net.dv8tion.jda.api.hooks.InterfacedEventManager.handle(InterfacedEventManager.java:96)
	at net.dv8tion.jda.internal.hooks.EventManagerProxy.handleInternally(EventManagerProxy.java:88)
	at net.dv8tion.jda.internal.hooks.EventManagerProxy.handle(EventManagerProxy.java:70)
[...]

Issue #945, opened 2021-10-03
[Feature Request] GraalVM (or different engine) for eval in Java 15+


MichailiK:
### Is your feature request related to a problem? Please describe.

The eval command is broken in Java 15 and above, as [Nashorn has been removed since Java 15](https://openjdk.java.net/jeps/372), which the MusicBot relies on.

Aside from powerusers, the eval command is currently the only way to list & leave guilds (servers).

### What is your ideal solution to the problem?

Add a new JavaScript Engine as a dependency, for example Graal. I haven't looked much into Graal. I'm not sure if it supports bindings or if it works in Java 8.

### How would this feature be used?

N/A

### Checklist

- [X] I have checked [the documentation](https://jmusicbot.com/) to make sure this feature doesn't already exist
- [X] I have searched for [similar feature requests](https://github.com/jagrosh/MusicBot/labels/enhancement)
- [X] I have confirmed that my requests is **not** on the [list of things that will not be added](https://jmusicbot.com/things-that-wont-be-added)
- [X] I am running the latest version of the bot: ![Release](https://img.shields.io/github/release/jagrosh/MusicBot.svg)

jagrosh:
what about changing the `eval = ` part of the config to a string, and let users who know what they are doing to specify their own interpreter

PR #1530, opened 2024-03-30
Custom Eval Engine

jagrosh:
### This pull request...
  - [X] Fixes a bug

### Description
Allows setting the eval engine in the config, and adds an error message if the provided engine does not exist

### Purpose
Later versions of Java don't include the Nashorn engine; this PR allows setting the engine to be used.

### Relevant Issue(s)
This PR closes issue #1501, #945

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
+++ b/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
@@ -40,7 +40,8 @@ public class BotConfig
     
     private Path path = null;
     private String token, prefix, altprefix, helpWord, playlistsFolder,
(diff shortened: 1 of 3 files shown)


jagrosh:
This is technically good-to-go but I feel like I want to change how we do things in the config just a little bit:
Instead of:
```
eval = true
evalengine = "Nashorn"
```
I'd like to do:
```
eval {
  enabled = true
  engine = "Nashorn"
}
```
However, this would cause backwards-compatibility issues that I'm not immediately sure how to solve. We'd need to investigate how HOCON handles overrides when the data type changes entirely.

jagrosh:
I'm going to opt to continue with how we currently do things, without nesting. I still think we might want to make this change later, but maybe all at once as part of a 'config update'

Issue #49, opened 2017-07-22
600 song limit


oyath:
Is this a hard limit on the bot or something, my autoplaylist only pulls from the first 600 of a youtube playlist i've created and any time i queue the playlist it only adds the first 600 songs. Any way to get this increased? Also searching for songs doesn't work it seems

jagrosh:
The 600 limit is enforced by an underlying library; I think I can remove that. As for searching, that's a known bug with the underlying library and I can't do anything to fix it until the library has a solution; in the meantime you can just restart the bot and the searching issue should go away.

Definitemars:
Hi there, 

Just updated my bot to 0.4.0 today and i'm finding that it caps out at 560 songs on youtube playlists. Have there been any fixes since the above question? Thank you.

Juustar:
> Hi there,
> 
> Just updated my bot to 0.4.0 today and i'm finding that it caps out at 560 songs on youtube playlists. Have there been any fixes since the above question? Thank you.

having the same issue here. had several youtube playlists at exactly 999 songs to avoid going over the old limit, but now i cant load the entire playlist. would prefer to find a fix than have to remake all my playlists

jagrosh:
We switched to a different fork of lavaplayer for this update; I'll investigate what may have changed and see if I can resolve it for 0.4.1

jagrosh:
Until then (or just for anyone wanting to make extremely-large playlists), you could make multiple youtube playlists and put them in a local playlist file.

jagrosh:
Will be a settable value in 0.4.1

PR #1542, opened 2024-04-10
Change playlist page count

jagrosh:
### This pull request...
  - [X] Fixes a bug

### Description
Increases playlist page count for YouTube playlists. Maybe this should be configurable? Is it worth making it configurable, or just set it to a huge number?

### Purpose
Allows longer playlists

### Relevant Issue(s)
This PR closes issue #49


jagrosh:
Note: need to check that this works, and if it does, we should consider making this a config item instead of a hard-coded value

Issue #519, opened 2020-10-19
[Feature Request] Deafen itself


kubo6472:
**Is your feature request related to a problem? Please describe.**
Bandwidth saving

**Describe the solution you'd like**
Adding config option to deafen itself.
e.g. https://groovy.zendesk.com/hc/en-us/articles/360042923932-Why-does-Groovy-deafen-itself-

**Describe the basic flow/steps of using this feature**
How this feature would be used:
1. boolean config option: deafen_itself=true/false
2. bot would deafen itself to save bandwidth
3. default to true

**Additional context**
Config option

**Have you confirmed that this idea is not already described here?**
- [x] https://github.com/jagrosh/MusicBot/projects/1

jagrosh:
Unless Discord's API has changed recently, that article is not accurate and deafening is performed clientside, so there is no change in bandwidth (bots "deafening" is purely a visual change).

kubo6472:
Afaik at least the server deafening voice state can be performed by the bot itself. Tried looking, but haven't found anything to back their claims, so I asked...

jagrosh:
If they or anyone else can share documentation provided by Discord that this is the case, I'd be happy to reopen this issue.

kubo6472:
Deal! Thank you for your time at least :)

kubo6472:
I asked the discord support

> Hi, please, I'd like some "official" or direct statement, that this musicbot server deafening itself does actually have any effect on the bandwidth it uses: https://groovy.zendesk.com/hc/en-us/articles/360042923932-Why-does-Groovy-deafen-itself-

Official response from their engineering team:

> Hello,
> 
>  thank you for your patience.
> 
>  To answer your question: Yes, it has an impact and helps us save bandwidth in voice channels.
> 
>  Unfortunately, we cannot provide you with any more information.  But please let us know if you have any further questions about the app!  :)

Ignaciox:
can +1 this. It actually does save bandwidth  @jagrosh

PR #1545, opened 2024-04-12
Add config option whether the bot should deafen itself

MichailiK:
### This pull request...
  - [ ] Fixes a bug

### Description
Adds a config option which allows users to configure whether the bot should deafen itself or not.

### Purpose
Although it is better for the bot to be deafened due to network bandwidth savings & increased privacy, some people are annoyed by the deafened icon and would prefer to disable deafen.

### Relevant Issue(s)
N/A, i remember there were a few complaints about this on the Discord server.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
+++ b/src/main/java/com/jagrosh/jmusicbot/BotConfig.java
@@ -41,7 +41,7 @@ public class BotConfig
     private Path path = null;
     private String token, prefix, altprefix, helpWord, playlistsFolder, logLevel,
             successEmoji, warningEmoji, errorEmoji, loadingEmoji, searchingEmoji;
-    private boolean stayInChannel, songInGame, npImages, updatealerts, useEval, dbots;
+    private boolean stayInChannel, deafen, songInGame, npImages, updatealerts, useEval, dbots;
     private long owner, maxSeconds, aloneTimeUntilStop;
     private double skipratio;
     private OnlineStatus status;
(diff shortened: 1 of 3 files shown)


jagrosh:
This looks good in terms of code. However:
1. Since we know we are working on attenuation, we need to make sure that when we do that we add a note to this feature's config description.
2. Do we have hard evidence that deafening actually reduces network load? At the time that JMusicBot was originally written, deafening was purely visual and didn't actually prevent receiving voice data (bots would still get it and could still 'record' audio while deafened, and the official Discord client effectively just silenced incoming audio). I hope this has changed, but is there an actual test anywhere proving that it has?

(note that I'm in favor of this change regardless of how we handle deafening; I just still haven't seen any hard evidence that it is anything more than a visual change)

MichailiK:
1. Mmh, the issue with attenuation is there seems to be a JDA 4 bug that only fires the speaking event once, then never again, so it cannot be implemented atm, at least without upgrading to JDA 5 im guessing.
2. Somewhat of a rant incoming:

You know, it would be awesome if [the response of Discord's support team](https://github.com/jagrosh/MusicBot/issues/519#issuecomment-716128114) were correct:
> Hello,
>
> thank you for your patience.
>
> To answer your question: Yes, it has an impact and helps us save bandwidth in voice channels.
>
> Unfortunately, we cannot provide you with any more information. But please let us know if you have any further questions about the app! :) 
    
Well, unfortunately, it looks like that's not actually the case & deafening yourself is not enough. You'll have to server deafen in order to not receive any traffic:

https://github.com/jagrosh/MusicBot/assets/39029839/1059a81a-ab63-4d0a-b521-42ceba86d2d7

(Notice how Network rx goes up as I speak, until i server deafen the bot)

That is... a pretty disappointing discovery how the support rep has blundered & actually only server deafening saves bandwidth...

MichailiK:
"Superseded" by #1551 following this discovery.

PR #1491, opened 2024-03-02
Always self-deafen

MichailiK:
### This pull request...
  - [ ] Fixes a bug
  - [ ] Introduces a new feature
  - [x] Improves an existing feature
  - [ ] Boosts code quality or performance

### Description
Self-deafen when joining voice channels.

### Purpose
Saves bandwidth & makes people less paranoid over the bot listening to them.

### Relevant Issue(s)
Resolves #519

PR #1551, opened 2024-04-22
Revert "Always self-deafen (#1491)"

MichailiK:
This PR reverts the always self-deafen. [An earlier comment of mine](https://github.com/jagrosh/MusicBot/pull/1545#issuecomment-2057789129) has proven that self-deafening doesn't actually help:

<hr/>

You know, it would be awesome if [the response of Discord's support team](https://github.com/jagrosh/MusicBot/issues/519#issuecomment-716128114) were correct:
> Hello,
>
> thank you for your patience.
>
> To answer your question: Yes, it has an impact and helps us save bandwidth in voice channels.
>
> Unfortunately, we cannot provide you with any more information. But please let us know if you have any further questions about the app! :) 
    
Well, unfortunately, it looks like that's not actually the case & deafening yourself is not enough. You'll have to server deafen in order to not receive any traffic:

https://github.com/jagrosh/MusicBot/assets/39029839/1059a81a-ab63-4d0a-b521-42ceba86d2d7

(Notice how Network rx goes up as I speak, until i server deafen the bot)

That is... a pretty disappointing discovery how the support rep has blundered & actually only server deafening saves bandwidth...

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
+++ b/src/main/java/com/jagrosh/jmusicbot/commands/MusicCommand.java
@@ -86,7 +86,6 @@ public abstract class MusicCommand extends Command
                 try 
                 {
                     event.getGuild().getAudioManager().openAudioConnection(userState.getChannel());
-                    event.getGuild().getAudioManager().setSelfDeafened(true);
                 }
                 catch(PermissionException ex) 
                 {


andy01745:
The bot is just staying 
image

andy01745:
Now will not work at all

MichailiK:
Small update in regards to this PR & situation: You can actually stop receiving audio from other users using the undocumented "Media Sink Wants" operation in VCs. It doesn’t look like JDA has implemented this though & thus we still receive audio, the deafen is just visual.
I’ll see some time soon if the Media Sink Wants opcode could get implemented into JDA.

VanillaChan6571:
Or... just make the bot server deafen itself? Why revert the self deafen? Is there no tag/api to server defean itself?

MichailiK:
2 things:
1. I believe it would cause confusion when JMusicBot deafens itself. It may require additional setup by server admins, by either granting the bot Deafen Members permissions, or server-deafening the bot themselves.
2. I've recently been informed that bots *can* stop receiving audio, but that this is left undocumented by Discord & has only been found thanks to reverse-engineering the Discord client. This "method" will need to be implemented in JDA, so I'm going to see if I can submit a PR for that some time soon, then update JMusicBot to make use of it.

PR #1552, opened 2024-04-22
Switch to lavalink's new YouTube audio source manager

MichailiK:
The lavalink devs seem to be separating the YouTube audio source manager into its own library: https://github.com/lavalink-devs/youtube-source. The new source manager seems to be more resilient against YouTube playback breaking.

This PR:

- switches to this new audio source manager
- removes the `AudioSourceManagers.registerRemoteSources(this);` call and instead registers each audio source manager lavaplayer provides individually, EXCEPT for the YouTube one
  - looking at lavaplayer's code, it doesn't look like you can exclude specific source managers, so to exclude the lavaplayer's built-in YouTube source manager, we have to register each source manager individually :/
- reverts to a stable version of lavaplayer
- gets rid of the logback filter we made to hide the "no google account provided thus cant play age restricted videos" log message, as its not present in that source manager.


ALocked1:
This change makes it so that you are required to use the full YouTube playlist link rather than just the playlist ID. If your local playlist txt files are still using just the ID, **they will not work anymore**; you have to add **https://youtube.com/playlist?list=** to the front of the ID. 
image

MichailiK:
I have asked the lavalink devs about that issue just now. They very graciously & rapidly provided a fix for this! I've updated the dependency to its latest version again.

IngwiePhoenix:
LGTM!

Not a Java-pro or alike, so I can not contribute much - except say that the previously mentioned fork does indeed work.
Whwn can we expect a merge?

CyberFlameGO (approved):

PR #1727, opened 2024-10-18
QoL Dev Build/Snapshot Builds Disable DM Bot.

VanillaChan6571:
### This pull request...
  - [-] Fixes a bug
  - [~] Introduces a new feature
  - [-] Improves an existing feature
  - [✓] Boosts code quality or performance

### Description
Disables Bot DMing Owner Version Annoyance when Dealing with Snapshot builds / Dev Builds.

### Purpose
When Dealing with Constant Stop and Starts. The Bot will always send a message on Ready which is annoying. So this checks if the build is  = Snapshot then dont send a dm.

### Relevant Issue(s)
Unknown.

### DIFF
--- a/src/main/java/com/jagrosh/jmusicbot/Listener.java
+++ b/src/main/java/com/jagrosh/jmusicbot/Listener.java
@@ -71,17 +71,22 @@ public class Listener extends ListenerAdapter
         });
         if(bot.getConfig().useUpdateAlerts())
         {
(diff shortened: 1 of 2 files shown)


MichailiK:
You can also set `updatealerts=false` in the config.txt, which is what I use & I believe jagrosh uses as well to stop the regular update alerts.

PR #1757, opened 2024-11-12
Added the ability to use own client-id and secret

Mayiaz:
### This pull request...
  - [x] Fixes a bug

### Description
Added the ability to use own google client-id and secret

### Purpose
I got an error while searching for music and fixed it by generating my own client id and secret and replacing them in the code in a hacky way


MichailiK:
Using your own application instead of YouTube on TV would very unlikely allowed you to actually play the tracks.

As the OAuth2 workaround has been patched by YouTube anyway, I currently have different plans on getting around & also hopefully preventing future IP blocks.

